Issue #2, opened 2016-03-30
Cannot save record with asset


bensonby:
> TypeError: Cannot read property '_presaveAssetTask' of undefined

- This is a regression
- It happens after batch save feature is introduced

PR #3, opened 2016-03-30
Fix asset save

bensonby:
connects #2

Calling on method `_presave` needs to be bound (`_presave` will call `self._presaveAssetTask`)


ben181231:
Picked as 7daec6e

PR #6, opened 2016-04-13
Fix wrong representation of public access

ben181231:
connect oursky/skygear-SDK-JS#156

### DIFF
--- a/lib/acl.js
+++ b/lib/acl.js
@@ -17,14 +17,25 @@ import _ from 'lodash';
 import Role from './role';
 
 export const AccessLevel = {
-  ReadLevel: 'read',
-  WriteLevel: 'write'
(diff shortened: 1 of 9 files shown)


lib/acl.js
```diff
@@ -112,31 +145,40 @@ export default class ACL {
   }
 
   removeWriteAccess(role) {
-    this.entries = _.filter(this.entries, function (perEntry) {
+    this.entries = _.filter(this.entries, (perEntry) => {
       return perEntry.level !== AccessLevel.WriteLevel ||
+        !perEntry.role ||
         perEntry.role.name !== role.name;
     });
   }
```
rickmak: What will happen if a user have write access, than called remote write access? he have read access left, or have no access?

PR #11, opened 2016-05-18
Implement user based ACL

yulapshun:


rickmak:
In order to merge this PR into master. I will add the following:

1. add usage/example.html
1. get gird of the everywhere eslint-disable-line
1. Return fast on checking permission

Issue #12, opened 2016-05-20
Support session store


carmenlau:
For browsers which don't support local storage, for example safari private mode.

rickmak:
Closed via 9db3cd7

chpapa:
Multiple requests for supporting login at Safari Private Mode, re-open this issue for investigation / discussion

PR #14, opened 2016-05-23
Support safari private mode

carmenlau:
connect #12 

Fallback to use cookie instead of localforage when local storage is not support.


rickmak:
Merged via 9db3cd7

PR #17, opened 2016-06-16
Build minified js file

cheungpat:
This will build `skygear.min.js` and the source map in the `dist` directory.
These files are uploaded to npm when the module is published

connects #16


cheungpat:
See if this is what you have in mind @rickmak

rickmak:
Merged via e716ca3

PR #30, opened 2016-07-18
#29 Support Batch Atomic saving

waynux:
connect #29


ben181231:
@waynux Could you add some test cases for that ?

waynux:
@ben181231 just add some test cases. thanks

ben181231:
Picked as 8bd8a7c

PR #31, opened 2016-07-19
Use memory localstorage and save as dependencies

rickmak:
It will enable skygear to run in node.
connects #80


ben181231:
Picked as 7bc2537

PR #36, opened 2016-07-20
Make skygear easier to include in a webpack project

rickmak:
localforage to 1.4.2, it support es6 and webpack since 1.3.3
Use browser WebSocket if websocket is not install, it allow webpack to
configure to use browser WebSocket if set 'websocket' as external.


ben181231:
Picked as 5c18812

Issue #22, opened 2016-06-28
Update and Delete API is confusing


TakLee96:
To Update:
``` javascript
let query = new skygear.Query(Note);
query.equalTo('_id', '123456...');

skygear.publicDB.query(query)
.then((records) => {
  let note = records[0];
  note['content'] = 'Hello New World';
  return skygear.publicDB.save(note);
}).then(...);
```

To Delete:
``` javascript
skygear.publicDB.delete({
  'id': 'note/123456...'
}).then(...);
```

It is confusing that Update API uses `_id` but delete API uses `id`. Also, if I try to use `id` for update, then there is 500 internal server error saying column `note.id` does not exist; if I try to use `_id` for delete, there is 400 bad request error saying `_id` should be of format '{type}/{id}' got "".

cheungpat:
You don’t need to put an emphasis on the word “confusing” to convince me that this is confusing.

The intended usage for your illustrated scenario should be this:

```javascript
let query = new skygear.Query(Note);
query.equalTo('_id', '123456...');

// for update
skygear.publicDB.query(query)
.then((records) => {
  let note = records[0]; // assuming there are more than one records, ditto for below
  note['content'] = 'Hello New World';
  return skygear.publicDB.save(note);
}).then(...);

// for delete
skygear.publicDB.query(query)
.then((records) => {
  let note = records[0];
  return skygear.publicDB.delete(note);
}).then(...);
```

To bring feature parity among SDKs, I think JS need a function to fetch a record by ID, if such function exists:

```javascript
skygear.publicDB.fetchByIds(["note/123456..."])
.then((records) => {
  // do work here
})
```

This way, the user does not need to know the subtlety between `_id` (which is a field in the record dictionary) and `id` (which is a property in the JavaScript object).

We also need a `deleteByIds` function so that the developer does not need to create an Object just to delete a Record if the RecordID is known already, as in your `delete` example.

--

If you encounter a 500 error, please open a new issue with the way you access the server (sample code), and the details of the error you got from the server.

We might also want to consider warning the user if the passed in object in `delete` and `save` function is not of the expected type, but admittedly I am not familiar with JavaScript convention in this particular scenario.

TakLee96:
`fetchByIds` and `deleteByIds` sound like a good solution.

``` javascript
const Note = skygear.Record.extend("note");
var query = new skygear.Query(Note);
[...]

Issue #4, opened 2016-04-01
Convenient method for get record by ID


rickmak:
We need to construct a query for that purpose now

TakLee96:
iOS SDK can do that, so I think it's reasonable to build one for JS SDK as well

PR #40, opened 2016-07-27
Add fetchById method to database

TakLee96:
connects #22 
connects #4

### DIFF
--- a/lib/database.js
+++ b/lib/database.js
@@ -17,6 +17,8 @@ const _ = require('lodash');
 
 import Cache from './cache';
 import Asset from './asset';
+import Record from './record';
+import Query from './query';
 import QueryResult from './query_result';
 
 export default class Database {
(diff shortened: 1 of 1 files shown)


lib/database.js
```diff
@@ -30,6 +32,11 @@ export default class Database {
     this._cacheStore = new Cache(this.dbID);
   }
 
+  fetchById(id) {
```
rickmak: Let name it getByRecordID. with reference to `getUsersByEmail`.

lib/database.js
```diff
@@ -30,6 +32,13 @@ export default class Database {
...
+    let [recordType, recordId] = id.split('/');
```
cheungpat: use `Record.parseID`

lib/database.js
TakLee96: @rickmak How about `getRecordByID`?

cheungpat:
added #4 to the pull request

rickmak:
Merge via 15ffc8a

Issue #10, opened 2016-05-16
Add url to asset when serialize to JSON


yulapshun:
Currently formJSON() of asset has url, but toJSON() remove the url. So a deserialize asset cannot be serialized to its original form.

rickmak:
The intention of trim the url at `toJSON` is because of `skygear-server` don't accept asset with `url`. Since it set the wrong expectation to user, i.e. the server can download a remote image url.

The usage of `toJSON` for application caching is completely valid. And the SDK should try our best to return identical object when the user calling `fromJSON`.

The solution will be move the trim logic to the actual API call, not couple with the skygear record.

PR #41, opened 2016-07-28
Make the asset toJSON and fromJSON mirror on $url key

rickmak:
Since skygear-server deserialization is ignoring extra key ($url here),
it is safe to include it in the record playload.
connects  #10


ben181231:
Picked as 15ffc8a

Issue #32, opened 2016-07-20
Lookup user by username


TakLee96:
We have `skygear.getUsersByEmail([ ... ])`, then why not `skygear.getUsersByUsername([ ... ])`?

cheungpat:
@rickmak How to query user record by email in JS SDK?

cheungpat:
The `getUsersByEmail` should be changed so that it return `Record` instead of `User`. Alternatively we can create a `discoverUsersByEmail` that will return `Record`.

PR #42, opened 2016-07-28
Support user discovery with username

cheungpat:
connects SkygearIO/skygear-server#90
connects SkygearIO/skygear-SDK-JS#32


ben181231:
Picked as e23b710

Issue #45, opened 2016-08-08
Query > Pagination: The preference on `page` was ignored.


cpryanpang:
According to the docs https://docs.skygear.io/js/guide/query#pagination-ordering, we can perform a query with paginated results by specifying either `query.page` or `query.offset`.

However, by real experiment, only `offset` worked; `page` never worked. We are pretty sure we didn't specifying both of them at the same time.

By a quick search at the code here: https://github.com/SkygearIO/skygear-SDK-JS/blob/master/lib/query.js#L38 (Not sure whether I am looking at the correct file 👻 )
I guess only the preference `offset` is effective; `page` is ignored.

Though, such issue does not bother our development much as we can still use `.offset` in this way:
`query.offset = limit * (page - 1)`

PR #47, opened 2016-08-09
Support `page` property on query

ben181231:
connect #45


rickmak:
@ben181231 Add a control at example for demo?

ben181231:
> Add a control at example for demo?
Added.

PR #54, opened 2016-08-22
Add API to get current user from server

ben181231:
connect SkygearIO/skygear-server#111

### DIFF
--- a/lib/container.js
+++ b/lib/container.js
@@ -181,6 +181,16 @@ export default class Container {
     });
   }
 
+  whoami() {
+    let container = this;
+    return container.makeRequest('me', {})
+    .then(function (res) {
+      return new container.User(res.result);
+    }, function (err) {
+      return Promise.reject(err);
+    });
+  }
+
   changePassword(oldPassword, newPassword, invalidate = false) {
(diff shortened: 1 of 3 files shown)


lib/container.js
```diff
@@ -181,6 +181,16 @@ export default class Container {
...
   }
 
+  whoami() {
+    let container = this;
+    return container.makeRequest('me', {})
+    .then(function (res) {
+      return new container.User(res.result);
+    }, function (err) {
+      return Promise.reject(err);
+    });
```
rickmak: I think `whoami` will fetch the remote user AND update the `_user` (i.e. keep the `_currentUser` in sync with remote.

ben181231: Do you expect the `me` request will return the access token ? 

Currently, the `access_token` field is a required field in the view of SDK, but the `me` endpoint will not return it. Should I inject the token back the returned object and set it to `_currentUser` ?

rickmak: Yes, let do this. It can also act as a access_token rotation in future.

ben181231:
@rickmak The PR updated.

Issue #33, opened 2016-07-20
Change username


TakLee96:
Users can change their email by calling `skygear.saveUser`. Why can't they change their username?

PR #63, opened 2016-09-08
Update username of a user in saveUser

cheungpat:
The saveUser function now sends new username to the server
when it is specified.

connects #33


rickmak:
Merge via 0f548e4

Issue #61, opened 2016-09-05
Support sign up anonymous user


royuen:

cheungpat:
@rickmak not sure if this is match what you have in mind: I will add `signupAnonymously()` taking no arguments which will send a `auth:signup` request with `username`, `email` and `password` being `null`.

This matches the behaviour of the iOS SDK.

cheungpat:
@rickmak said need to allow the modifying username of a user in JS SDK.

PR #64, opened 2016-09-08
Allow sign up anonymously

cheungpat:
This implements signupAnonymously which creates a user account
without email address and username.

connects #61


rickmak:
Merge via 22cc1d0

Issue #73, opened 2016-09-26
Introduce copy constructor for record


louischan-oursky:
Record is not currently copy constructible because of #72

rickmak:
@louischan-oursky please check https://github.com/SkygearIO/skygear-SDK-JS/pull/95 I modify to support acl too.

Issue #72, opened 2016-09-26
Disambiguate the meaning of `_id`


louischan-oursky:
When `_id` is included as a property of the `attrs` of `Record` constructor, it is expected to be a typed id, i.e. `my_table_name/some-id`

After the record instance is constructed, `_id` is typeless id, i.e. `some-id`.

It is very confusing. It would be better to support `id` as one of the property of `attrs` which is interpreted as typed id. The new constructor must issue deprecation warning for users who still use `_id` for the constructor. Then `_id` will always mean typeless id.

```
const Note = skygear.Record.extend('note');
const noteRecord = new Note({
  _id: 'note/1',
});
console.log(noteRecord._id); // this logs "1"
```

cheungpat:
@louischan-oursky 

We have this terminology: Record ID (`note/42`) consists of Record Type (`note`) and Record Name (`42`).

The `_id` field contains record type information because this is how the data is transmitted from the server to the SDK. The data is passed to the constructor you mentioned so to construct a `Record`. Because of this, in the forseeable future, `_id` passed to this constructor will have type information because not doing so will break HTTP API compatibility.

In the realm of the JS SDK, I think the recommended approach is to get Record ID information from `id` property (without `_`), which returns record type + record name. You are not supposed to use or know about `_id`. For this reason there is no point in “disambiguating the meaning of `_id`”.

You are having problem because you are unable to pass a `Record` into the constructor to create a new object (i.e. the copy constructor you referenced in #73). To solve this problem is easy: if `_id` does not contain record type information, the constructor should assume that the record type is the same as the record type when calling `extend(...)`.

cheungpat:
This should be handled in #73

PR #74, opened 2016-09-26
refs #72 #73 Accept typeless _id in constructor

louischan-oursky:


oursky-travis:
@louischan-oursky, thanks for your PR! By analyzing the annotation information on this pull request, we identified @rickmak to be a potential reviewer

lib/record.js
```diff
@@ -92,11 +92,20 @@ export default class Record {
...
       id = uuid.v4();
     } else {
-      let [type, name] = Record.parseID(id);
-      if (type !== this._recordType) {
-        throw new Error('_id is not valid. RecordType mismatch.');
+      let tuple = null;
+      try {
+        tuple = Record.parseID(id);
+      } catch (e) {
+        // id is typeless
```
cheungpat: Having an exception when parsing the ID doesn’t necessarily imply the string has no type information.

lib/record.js
```diff
@@ -92,11 +92,20 @@ export default class Record {
...
+        // default to recordType
+      }
+      if (tuple) {
+        let [type, name] = tuple;
+        if (type !== this._recordType) {
```
cheungpat: Perhaps you also want to check if `_recordType` exists, and that it equals `this._recordType`.

lib/record.js
cheungpat: I wouldn’t say no if you don’t call `parseID` when both `_recordType` field and `_id` exists.

test/record.js
```diff
@@ -85,6 +85,34 @@ describe('Extended Record', function () {
...
+  it('reject _id with different type', function () {
+    expect(function () {
+      let r = new Note({
+        _id: 'box/2'
+      });
+    }).to.throw(
+      '_id is not valid. RecordType mismatch.'
+    );
+  });
+
```
cheungpat: You need test case here for the cases:

1. When `_recordType` matches
2. When `_recordType` does not matches

cheungpat:
I wish @louischan-oursky could step up and make changes to the PR. We could all benefit from this PR here.

rickmak:
Superset by  #95

Issue #77, opened 2016-09-28
Add getUsersByUsername


rickmak:
Same as `getUsersByEmail`, intended for admin query.

PR #78, opened 2016-09-28
Add getUsersByUsername to container 

rickmak:
connects #77

### DIFF
--- a/lib/container.js
+++ b/lib/container.js
@@ -240,11 +240,12 @@ export default class Container {
     });
   }
 
-  getUsersByEmail(emails) {
+  _getUsersBy(emails, usernames) {
     let container = this;
     return new Promise(function (resolve, reject) {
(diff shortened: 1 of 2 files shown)


lib/container.js
```diff
@@ -240,11 +240,12 @@ export default class Container {
...
   }
 
-  getUsersByEmail(emails) {
+  _getUsersBy(emails, usernames) {
     let container = this;
     return new Promise(function (resolve, reject) {
       container.makeRequest('user:query', {
-        emails: emails
+        emails: emails,
+        usernames: usernames
```
ben181231: Can server handle the request when both `emails` and `usernames` are set ?

rickmak: Yes: https://github.com/SkygearIO/skygear-server/blob/master/pkg/server/skydb/pq/user.go#L264

Issue #91, opened 2016-10-27
changePassword doesn't update SDK access token


akiroz:
changePassword doesn't update SDK access token

- Skygear SDK Date/Version: `0.17`
- JS runtime: `Firefox 49`
- Skygear Server Date/Version: `Skygear Cloud Oct.27 2016`

#### Expected Results
access token updates after a password change

#### Actual Results
access token didn't update after a password change causing later requests to be Unauthorized

#### Steps to reproduce
1. `skygear.loginWithUsername`
2. `skygear.changePassword`
3. `skygear.whoami` will fail with HTTP 401

Issue #90, opened 2016-10-27
changePassword doesn't return User object as shown in the doc


akiroz:
`changePassword` doesn't return User object as shown in the doc

- Skygear SDK Date/Version: `0.17`
- JS runtime: `Firefox 49`
- Skygear Server Date/Version: `Skygear Cloud Oct.27 2016`

![screenshot from 2016-10-27 14-54-27](https://cloud.githubusercontent.com/assets/8231917/19759841/53391762-9c61-11e6-9278-c4d2e0ad7864.png)

#### Expected Results
`changePassword` returns a Promise that resolves to a User object 

#### Actual Results
`changePassword` returns a Promise that resolves to a string

#### Steps to reproduce
1. `skygear.loginWithUsername`
2. `skygear.changePassword`

PR #92, opened 2016-10-27
Fix change password not handling auth response

cheungpat:
The correct behavior is to handle the auth response (e.g. use new access
token) and to return the user object.

connects #90
connects #91


oursky-travis:

cheungpat:
@akiroz The fix should have been a simple one. Anyway you might want to check if this works for you.

Issue #86, opened 2016-10-26
Cloudcode registry


rickmak:
Implement all function for register and store it.

The function should be:
- before_save
- after_save
- before_delete
- after_delete
- handler
- op
- every

PR #99, opened 2016-11-03
Add cloud code registry

rickmak:
connect #86

### DIFF
--- a/lib/cloud/index.js
+++ b/lib/cloud/index.js
@@ -1,14 +1,25 @@
 import registry from './registry';
 import crypto from 'crypto';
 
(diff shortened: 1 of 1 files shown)


lib/cloud/index.js
```diff
@@ -0,0 +1,60 @@
+import registry from './registry';
+import crypto from 'crypto';
+
+function funcName(func) {
```
cheungpat: The name of the function should be preferrably consistent in subsequent invocation. Is it not possible?

rickmak: Let use registration order as suffix than.

PR #102, opened 2016-11-04
Add initial http transport implementation

rickmak:
- return the init to server
- call lambda
- call timer
- call hook

connects #87

### DIFF
--- a/lib/bin.js
+++ b/lib/bin.js
@@ -41,3 +41,12 @@ if (cmd === '--settings') {
 
 const codePath = path.join(process.cwd(), cmd);
 require(codePath);
+
+let transport;
+if (settings.http) {
+  transport = require('./cloud/transport/http');
+} else {
+  throw new Error('Currently, only http transport is supported.');
+}
+
+transport.start();
--- a/lib/cloud/transport/common.js
+++ b/lib/cloud/transport/common.js
@@ -1,0 +1,62 @@
+export default class HTTPTransport {
+  constructor(registry) {
+    this.registry = registry;
+  }
+
+  start() {
+    throw new Error('Not implemented');
+  }
+
+  initHandler(payload) {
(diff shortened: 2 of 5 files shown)


lib/bin.js
```diff
@@ -41,3 +41,12 @@ if (cmd === '--settings') {
 
 const codePath = path.join(process.cwd(), cmd);
 require(codePath);
+
+let transport;
+if (settings.http) {
+  transport = require('./cloud/transport/http');
+} else {
+  throw new Error('Currently, only http transport is support.');
```
cheungpat: s/support/supported/

lib/cloud/transport/common.js
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  hookHandler(payload) {
+    const {
+      name,
+      param
+    } = payload;
+    const func = this.registry.getFunc('hook', name);
+    if (!func) {
+      throw new Error('Databse hook not exist');
```
cheungpat: s/Databse hook not exist/Database hook does not exist/

lib/cloud/transport/http.js
```diff
@@ -0,0 +1,89 @@
...
+    this.server = http.createServer(this.readReq);
+    this.server.on('clientError', (err, socket) => {
+      socket.end('HTTP/1.1 400 Bad Request\r\n\r\n');
+    });
+    this.server.on('error', (err, socket) => {
+      socket.end('HTTP/1.1 500 Internal Server Error\r\n\r\n');
+    });
+    console.info('Listening on: ' + settings.httpAddr);
+    // TODO: parse the httpAddr
+    this.server.listen(9000, '0.0.0.0');
```
cheungpat: This need to read environment variable but I guess we could wait later.

Issue #103, opened 2016-11-07
Support AuthProvider in cloud


rickmak:
Follow up https://github.com/SkygearIO/skygear-SDK-JS/issues/88

PR #114, opened 2016-11-09
Js Auth provider in cloud code

rickmak:
connects #103


cheungpat:
@rickmak this PR need rebasing

PR #115, opened 2016-11-09
Support plugin event

ben181231:
connect #100

### DIFF
--- a/lib/cloud/registry.js
+++ b/lib/cloud/registry.js
@@ -74,14 +74,16 @@ export class Registry {
     this.funcMap.op[name] = func;
   }
 
-  registerEvent(name, func, options) {
-    const opts = {
-      ...options,
-      name
-    };
+  registerEvent(name, func) {
+    const eventParams = this.paramMap.event.filter((e) => e.name === name);
+    if (eventParams.length === 0) {
+      this._addParam('event', { name });
+    }
 
-    this._addParam('event', opts);
-    this.funcMap.event[name] = func;
+    const funcList = this.funcMap.event[name] || [];
+    funcList.push(func);
+
+    this.funcMap.event[name] = funcList;
   }
 
   registerTimer(name, func, options) {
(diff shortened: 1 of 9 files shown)


oursky-travis:
@ben181231, thanks for your PR! By analyzing the annotation information on this pull request, we identified @rickmak to be a potential reviewer

lib/cloud/registry.js
```diff
@@ -72,6 +74,16 @@ export class Registry {
...
   }
 
+  registerEvent(name, func, options) {
+    const opts = {
+      ...options,
+      name
+    };
+
+    this._addParam('event', opts);
+    this.funcMap.event[name] = func;
```
rickmak: We will need to support multiple func hook on to one event. So the name cannot just set according to the event name.

lib/cloud/transport/http.js
```diff
@@ -42,9 +43,17 @@ class HTTPTransport extends CommonTransport {
     this.server.on('error', (err, socket) => {
       socket.end('HTTP/1.1 500 Internal Server Error\r\n\r\n');
     });
-    console.info('Listening on: ' + settings.httpAddr);
-    // TODO: parse the httpAddr
-    this.server.listen(9000, '0.0.0.0');
+
```
rickmak: This split should belong to setting modules.

ben181231: Okay

lib/cloud/registry.js
ben181231: Oh...

Dockerfile-development
```diff
@@ -5,11 +5,20 @@ RUN \
...
+
+COPY . /usr/src/skygear
 
-COPY . /usr/src/app
-RUN (cd /usr/src/app; npm install && npm link)
+RUN \
+  cd /usr/src/skygear && \
+  npm install && \
+  npm run prepublish && \
+  npm install -g $PWD
```
cheungpat: This Dockerfile is aimed for skygear contributor, and the previous code allow development without rebuilding Docker image every step of the way

ben181231: I got that.

ben181231:
Updated

cheungpat:
@rickmak please merge this for release

Issue #109, opened 2016-11-08
Provide pg connection when calling database hook cloud code


cheungpat:
follows #89 and #105

PR #117, opened 2016-11-10
Provide pq.pool as the third argument on DB hook

rickmak:
connects #109

### DIFF
--- a/lib/cloud/transport/common.js
+++ b/lib/cloud/transport/common.js
@@ -20,6 +20,7 @@ import {
 import stream from 'stream';
 import {IncomingForm} from 'formidable';
 import {parse} from 'url';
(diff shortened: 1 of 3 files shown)


lib/cloud/transport/common.js
```diff
@@ -86,7 +87,7 @@ export default class CommonTransport {
     if (param.original) {
       originalRecord = new Record(_type, param.original);
     }
-    let record = func(incomingRecord, originalRecord);
+    let record = func(incomingRecord, originalRecord, pool);
```
cheungpat: On one hand, this is even better than py-skygear because the pg connection is only created when the connection is needed. But, on the other hand, I do not understand the importance since the user can just import the pool from the `pg` module.

PR #124, opened 2016-11-15
Expose pg pool, settings and CloudCodeContainer

cheungpat:
The `pg` and `settings` module are not exposed in skygear cloud package. Furthermore, cloud code need a container that can be instantiated and be configured for cloud code use (such as master key and impersonate another user).


oursky-travis:

Issue #97, opened 2016-11-02
Support static assets


royuen:
- Providing hook for declaring local files as static assets.
- Server static asset during develop
- Collect the local asset in cloud deployment

PR #125, opened 2016-11-15
Support static asset

rickmak:
- Providing `staticAsset` for declaring local files as static assets
- Server static asset during develop if `SERVE_STATIC_ASSETS` is true
- Provide `skyegar-asset` command to Collect local asset in cloud deployment

connects #97

### DIFF
--- a/asset.js
+++ b/asset.js
@@ -1,0 +1,2 @@
+#!/usr/bin/env node
+require('./dist/asset_collect.js');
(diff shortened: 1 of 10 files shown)


ben181231:
the `public_html` folder is not supported in this PR, right?

rickmak:
it is not included.

cheungpat:
seems okay

Issue #127, opened 2016-11-16
Incorrect error format


cheungpat:
If an error occurred with js cloud code, the skygear-node returns an error like this:
```json
{"error":"error message"}
```

skygear-server expects the data in `common.ExecError` format:
```json
{"error":{"name":"<error type>","code":10000,"message":"<error message>"}}
```

referencing issue at skygeario/skygear-server#232

PR #136, opened 2016-11-17
Return error to skygear-server in correct format

cheungpat:
connects #127

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -7,7 +7,7 @@
   "plugins": [
     "add-module-exports",
     ["transform-builtin-extend", {
-      "globals": ["Array"],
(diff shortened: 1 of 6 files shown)


oursky-travis:

ben181231:
@cheungpat Currently, the `lib/error.js` is imported to container as `ErrorCodes`. Can you also rename it since it is now exporting not only error codes.

cheungpat:
What do you think if I do this:

```js
export const ErrorCodes = { /* ... error codes and value here ... */ };
```

ben181231:
Seems great.

cheungpat:
updated

PR #138, opened 2016-11-18
cloud: Fix undefined result with no event handlers

cheungpat:
CommoonTransport resolve a promise with an undefined result if no
event handlers are defined. This will cause the HTTP transport
to write undefined to ServerResponse, which is not allowed.


oursky-travis:

ben181231:
Picked as d0a846a

PR #144, opened 2016-12-08
Fix not able to create Record with `id` attr

cheungpat:
The constructor of Record prefers `id` to pass record ID, but it is not
possible to use this attr as the update function will set the record ID
to `id` property, which is readonly.

Not having this change will result in TypeError:

```
TypeError: Cannot set property id of #<Record> which has only a getter
```

### DIFF
--- a/lib/record.js
+++ b/lib/record.js
@@ -92,6 +92,7 @@ export default class Record {
       }
       id = name;
     }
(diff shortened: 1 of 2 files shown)


oursky-travis:

rickmak:
Merge via 73e8773

PR #149, opened 2016-12-15
Fix binary static asset encoding

cheungpat:
The existing implementation cannot send binary static asset through
a handler. The new static asset handler returns a buffer so that
that the transport layer uses the buffer's base64 encoding to fix this problem.


oursky-travis:
@cheungpat, thanks for your PR! By analyzing the history of the files in this pull request, we identified @rickmak to be a potential reviewer.

rickmak:
Merge via 72ff6b8

PR #150, opened 2016-12-16
Support unregister device

ben181231:
connect SkygearIO/skygear-server#245
connect SkygearIO/skygear-server#249

### DIFF
--- a/lib/container.js
+++ b/lib/container.js
@@ -31,7 +31,7 @@ import Pubsub from './pubsub';
 import {RelationAction} from './relation';
 import Geolocation from './geolocation';
 import {Sequence} from './type';
-import {ErrorCodes} from './error';
+import {ErrorCodes, SkygearError} from './error';
 import {EventHandle} from './util';
 
 var React;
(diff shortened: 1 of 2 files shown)


oursky-travis:
@ben181231, thanks for your PR! By analyzing the history of the files in this pull request, we identified @rickmak and @cheungpat to be potential reviewers.

lib/container.js
```diff
@@ -164,8 +164,15 @@ export default class Container {
...
-    this.clearCache();
-    return container.makeRequest('auth:logout', {}).then(function () {
+    let unregisterDevicePromise = Promise.resolve();
+    if (this.deviceID) {
+      unregisterDevicePromise = this.unregisterDevice();
+    }
+    return unregisterDevicePromise.then(function () {
+      container.clearCache();
+      return container.makeRequest('auth:logout', {});
+    })
```
rickmak: I am more favour in following logic to make the handling change as part of fails handle. Not prior check. And this part of code is actually duplicate of `if (!this.deviceID) {` check in  `unregisterDevice`

```
unregisterDevicePromise.then(function () {
      container.clearCache();
      return container.makeRequest('auth:logout', {});
}, function() {
// Check if the Error is missing device id?
// If yes, logout. else throw it.
}).then(function () {...});
````

ben181231:
The PR is updated.

PR #156, opened 2017-01-04
Accept bundle name when register device

ben181231:
connect SkygearIO/skygear-server#239

### DIFF
--- a/lib/container.js
+++ b/lib/container.js
@@ -364,7 +364,15 @@ export default class Container {
     });
   }
 
-  registerDevice(token, type, bundleName) {
+  /**
+   * You can register your device for receiving push notifications.
+   *
+   * @param {string} token - The device token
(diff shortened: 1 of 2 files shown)


oursky-travis:
@ben181231, thanks for your PR! By analyzing the history of the files in this pull request, we identified @cheungpat and @rickmak to be potential reviewers.

lib/container.js
```diff
@@ -364,7 +364,7 @@ export default class Container {
     });
   }
 
-  registerDevice(token, type) {
+  registerDevice(token, type, bundleName) {
```
rickmak: To incrementally polish our class doc. It is better to add a short description on what is the bundleName parameter is for.

rickmak: Better to change the parameter to `topic` to align with the server naming. And mention the usual value is bundleName.

PR #169, opened 2017-02-13
Add timeout to container request

cheungpat:
This is a re-apply of the commit 1d3e13a after upgrading superagent.

connects SkygearIO/skygear-server#271


cheungpat:
It would be great if there is suggestion on how to write test case that will actually check the timeout, but that doesn’t seem supported with superagent-mock.

rickmak:
Merge via 0e81308

Issue #165, opened 2017-02-09
Support asset signer


royuen:
This is brought up from implementing the CMS plugin about supporting assets (https://github.com/oursky/skygear-cms/issues/124).

For assets it could have private assets which needs to sign the asset. 

It doesn't have this sign function in API. We would need to do it directly from JS SDK for plugin support.

PR #170, opened 2017-02-13
Add url signing function to cloud

b123400:
connect #165

### DIFF
--- a/lib/cloud/asset.js
+++ b/lib/cloud/asset.js
@@ -1,8 +1,14 @@
+import crypto from 'crypto';
 import fs from 'fs';
 import path from 'path';
 import mime from 'mime-types';
 import {SkygearResponse} from './transport/common';
 import registry from './registry';
+import {settings} from './settings';
+import URLSafeBase64 from 'urlsafe-base64';
+import S3URLSigner from 'amazon-s3-url-signer';
+import request from 'superagent';
+import URL from 'url';
 
 /**
  * staticAssetHandler — default handler for serving static assets with during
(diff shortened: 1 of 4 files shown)


rickmak:
This PR is fs only. I would prefer to merge with all 3 signer together. i.e. `fs`, `cloud` and `s3`

b123400:
Updated to support s3, cloud and fs

lib/cloud/asset.js
```diff
@@ -47,3 +53,150 @@ export function staticAssetHandler(req) {
...
+    const host = this.settings.cloudAssetHost;
+    const url = `${host}/token/${appName}`;
+
+    return superagent.get(url)
+      .accept('application/json')
+      .set('Authorization', `Bearer ${token}`)
+      .query({expired_at: expire.toString()})
+      .then(response => {
+        const body = response.body;
+        console.log('fuck', body);
```
rickmak: The choice of word is too strong.

lib/cloud/settings.js
```diff
@@ -78,5 +78,29 @@ export const settings = {
...
+  assetStoreURLPrefix: process.env.ASSET_STORE_URL_PREFIX ||
+    'http://127.0.0.1:3000/files',
+  assetStoreURLExpireDuration:
+    parseInt(process.env.ASSET_STORE_URL_EXPIRE_DURATION) || 3600,
+
+  // S3 asset settings
+  assetStoreS3AccessKey: process.env.ASSET_STORE_ACCESS_KEY || '',
+  assetStoreS3SecretKey: process.env.ASSET_STORE_SECRET_KEY || 'asset-secret',
+  assetStoreS3Bucket: process.env.ASSET_STORE_BUCKET || 'bucket-name',
+  assetStoreS3Region: process.env.ASSET_STORE_REGION || 'eu-west-1',
```
rickmak: Although we should not relay on default. I think the common default is `us-east-1`?

lib/cloud/asset.js
```diff
@@ -47,3 +53,150 @@ export function staticAssetHandler(req) {
     body: data
   });
 }
+
+class Signer {
+  sign() {
```
rickmak: The sign interface is designed to return a `Promise`? May be wee some comment on this.

lib/cloud/asset.js
```diff
@@ -47,3 +53,150 @@ export function staticAssetHandler(req) {
...
+      .update(appName)
+      .update(name)
+      .update(expired.toString())
+      .update(this.signerToken.extra)
+      .digest('base64');
+    const signature_and_extra =
+      encodeURIComponent(`${hash}.${this.signerToken.extra}`);
+
+    return `${prefix}/${appName}/${name}` +
+      `?expired_at=${expired}&signature=${signature_and_extra}`;
```
rickmak: This return should be `Promise`?

lib/cloud/asset.js
```diff
@@ -47,3 +53,150 @@ export function staticAssetHandler(req) {
...
+    const url = this.signer.getUrl('GET', name, this.bucket, duration);
+    return Promise.resolve(url);
+  }
+}
+/**
+ * Return a signer for the current configuration.
+ *
+ * @return {Signer}
+ */
+export function getSigner() {
```
rickmak: We only need one instance of signer. The name `getSigner` also imply this to me.

So I have two quesiton:
- What is the consideration on returning new instance on every `getSigner`?
- The name `signerFactory` will tell everyone that the new signer will return on every call.

lib/cloud/asset.js
```diff
@@ -47,3 +53,150 @@ export function staticAssetHandler(req) {
...
+    super();
+    this.settings = _settings;
+    this.signerToken = null;
+  }
+
+  refreshSignerToken() {
+    const appName = this.settings.appName;
+    const duration = this.settings.assetStoreURLExpireDuration;
+    const expire = Math.floor(Date.now() / 1000) + duration;
+    const token = settings.cloudAssetToken;
```
rickmak: `this.settings`

lib/cloud/asset.js
```diff
@@ -47,3 +53,150 @@ export function staticAssetHandler(req) {
...
+      .set('Authorization', `Bearer ${token}`)
+      .query({expired_at: expire.toString()})
+      .then(response => {
+        const body = response.body;
+        console.log('fuck', body);
+        this.signerToken = {
+          value: body.value,
+          expiredAt: Date.parse(body.expired_at),
+          extra: body.extra
+        };
```
rickmak: Just put those data in `this`, not `this.signerToken`? I think an object should not need two/three level of internal data structure to make things complex.

And flat structure it save typing ;)

lib/cloud/asset.js
```diff
@@ -47,3 +53,150 @@ export function staticAssetHandler(req) {
...
+      `?expired_at=${expired}&signature=${signature_and_extra}`;
+  }
+}
+
+class S3Signer extends Signer {
+  constructor(_settings) {
+    super();
+    this.settings = _settings;
+    const key = _settings.assetStoreS3AccessKey;
+    const secret = _settings.assetStoreS3SecretKey;
```
rickmak: Why not `this.settings`?

lib/cloud/asset.js
```diff
@@ -47,3 +53,150 @@ export function staticAssetHandler(req) {
...
+  constructor(_settings) {
+    super();
+    this.settings = _settings;
+  }
+
+  sign(name) {
+    const prefix = this.settings.assetStoreURLPrefix;
+    const duration = this.settings.assetStoreURLExpireDuration;
+    const expire = Math.floor(Date.now() / 1000) + duration;
+    const secret = this.settings.assetStoreSecret;
```
rickmak: If we assume settings won't change. Just copy the setting into the `this`, not querying everytime during sign?

lib/cloud/asset.js
b123400: Sorry this is totally debug code!!

PR #181, opened 2017-03-27
Accept arguments to create profile when creating user

b123400:
connect https://github.com/SkygearIO/skygear-server/issues/54

I guess there would be comments on how the API looks and I have to update few times anyway, so I won't work on other platforms until this is confirmed ok.

Added arguments to these 2 APIs.
```
singupWithUsername(username, password, {
  name: "some name"
}).then(([user, profile])=> { ... })
```
```
singupWithEmail(email, password, {
  name: "some name"
}).then(([user, profile])=> { ... })
```

And 2 new APIs:
```
singupWithUsernameAndProfiles(username, password, {
  "user": {
    name: "some name"
  },
  "private_user": {
    secret: ...
  }
}).then(([user, profile1, profile2])=> { ... })
```
```
singupWithEmailAndProfiles(email, password, {
  "user": {
    name: "some name"
  },
  "private_user": {
    secret: ...
  }
}).then(([user, profile1, profile2])=> { ... })
```

When a profile is supplied, the return value in promise becomes an array of `[user, ...profiles]`. This is because we are creating multiple records in database, and there is no clear relationship between `UserRecord` and `User`, I just want to follow the way it was, to separate them and therefore return them in an array. Any suggestion what should it be like?

### DIFF
--- a/lib/container.js
+++ b/lib/container.js
@@ -120,6 +120,20 @@ export default class Container {
     return this._signup(null, email, password);
   }
 
+  signupWithUsernameAndProfile(username, password, profile = {}) {
+    return this.signupWithUsername(username, password)
+    .then((user)=>
+      this._createProfile(user, profile)
+    );
+  }
+
+  signupWithEmailAndProfile(email, password, profile = {}) {
+    return this.signupWithEmail(email, password)
+    .then((user)=>
+      this._createProfile(user, profile)
+    );
+  }
+
   signupAnonymously() {
     return this._signup(null, null, null);
   }
@@ -133,6 +147,14 @@ export default class Container {
         password: password
       }).then(container._authResolve.bind(container)).then(resolve, reject);
     });
+  }
+
+  _createProfile(user, profile) {
+    let record = new this.UserRecord({
+      _id: 'user/' + user.id,
+      ...profile
+    });
+    return this.publicDB.save(record);
   }
 
   _authResolve(body) {
--- a/test/container.js
+++ b/test/container.js
@@ -349,6 +349,46 @@ describe('Container auth', function () {
       });
   });
 
+  it('should signup with profile successfully', function () {
+    return container
+      .signupWithUsernameAndProfile('username', 'passwd', {
+        'key': 'value'
+      })
+      .then(function (profile) {
+        assert.equal(
+          container.accessToken,
+          'uuid1');
+        assert.instanceOf(container.currentUser, container.User);
+        assert.equal(
+          container.currentUser.id,
+          'user:id1'
+        );
+        assert.equal(profile.key, 'value');
+      }, function () {
+        throw new Error('Signup failed');
+      });
+  });
+
+  it('should signup with email and profile successfully', function () {
+    return container
+      .signupWithEmailAndProfile('user@email.com', 'passwd', {
+        'key': 'value'
+      })
+      .then(function (profile) {
+        assert.equal(
+          container.accessToken,
+          'uuid1');
+        assert.instanceOf(container.currentUser, container.User);
+        assert.equal(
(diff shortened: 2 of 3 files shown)


oursky-travis:
@b123400, thanks for your PR! By analyzing the history of the files in this pull request, we identified @rickmak, @cheungpat and @ben181231 to be potential reviewers.

ben181231:
From the changes, I found the following behaviors:

1. Multiple **public** profiles are supported;
2. **No private** profiles are supported during sign up API;
3. The profile record types will follow the **keys** of the 3rd argument of `signupWithXXXAndProfiles()`, which will be specified by developers.

Do we are agree with that? If yes, this PR will be merged.

@rickmak @cheungpat

rickmak:
Related feature: https://github.com/SkygearIO/features/issues/48

2. What is the use case to support multiple public profile, while no private profile? How will the user interact with multiple public profile? I don't get it.
3. For all record type we call `Profile` or `User`. I think it is better to keep the ID same as the `_user`.

ben181231:
I think it is okay to have the hidden / undocumented feature to support multiple public profiles during signup. And we still have `signupWithEmail()` and `signupWithUsername()` supporting one public profile in user record.

rickmak:
Let drop the `singupWithEmailAndProfiles` before we have decide on SkygearIO/features#48.

For use-case we need multiple profile, we should also design how the user interact with `getUsersByUsername`. If we introduce the `private` user in this PR. We need to modify all user related API to reflect the change. Say we need to decided weather we return both public/private user attributes in `getUsersByUsername`...

b123400:
Updated to remove the functions that create multiple profiles

ben181231:
Just found that `signupWithXXX()` returning a promise resolving different types of objects depending on whether the caller is proving a profile. Can we align them into one, like `user` object? We may also query the user profile and inject it to `user` object.

lib/container.js
```diff
@@ -112,12 +112,28 @@ export default class Container {
...
-  signupWithEmail(email, password) {
-    return this._signup(null, email, password);
+  signupWithEmail(email, password, profile = null) {
+    const promise = this._signup(null, email, password);
+    if (profile === null) {
+      return promise;
+    }
+    return promise.then((user)=>
+      this._createProfile(user, profile).then(createdProfile =>
+        [user, createdProfile]
```
rickmak: Different return value type seem not a good idea. Agree?

`signupWithEmail(email, password)` will resolve to an object, while `signupWithEmail(email, password, {})` will resolved to an array of object.

Since we want an enchantment, not a breaking change here. Probably we don't want break existing code using `signupWithEmail`.

So a sensible solution will be 
`signupWithEmailAndProfile(email, password, {})` will resolved to an array of object.

lib/container.js
```diff
@@ -135,6 +151,14 @@ export default class Container {
     });
   }
 
+  _createProfile(user, profile) {
+    let record = new this.UserRecord({
+      _id: 'user/' + user.id,
+      ...profile
+    });
+    return this.publicDB.save(record);
+  }
```
rickmak: I just discover a `saveUser` which is confusing with Profiles....

ben181231: `saveUser` is saving a user object instead of a user profile ?

lib/container.js
cheungpat: In my opinion, the utility of the `_user` object is pretty low. How about returning a profile from `signupWithEmailAndProfile` instead of returning an array of object?

lib/container.js
rickmak: `saveUser` accepting user object which is intended for update of username, email and user role.

b123400:
Updated to only return the profile object

PR #182, opened 2017-03-30
Add an function to return user-readable error message

b123400:
connect https://github.com/SkygearIO/skygear-SDK-Android/issues/89

This PR adds a function that returns user-friendly message for error.
This PR does not do localisation, any suggestion how (like, which library?) should we do localisation in JS?

Update: localisation will be done in another issue, only user-friendly message for this PR

### DIFF
--- a/lib/error.js
+++ b/lib/error.js
@@ -75,6 +75,52 @@ export class SkygearError extends Error {
     return `SkygearError: ${this.message}`;
   }
 
+  /* eslint-disable complexity */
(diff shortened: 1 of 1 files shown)


cheungpat:
@rickmak suggestion for localization library for js

rickmak:
Use this? https://formatjs.io by yahoo.

rickmak:
Merge via ba8a377

PR #184, opened 2017-04-06
Add pubsub.once helper function

akiroz:
Useful for working with ES7 async/await:
```js
async function foo() {
  await skygear.pubsub.once('bar');
  // ...
}
```

### DIFF
--- a/lib/pubsub.js
+++ b/lib/pubsub.js
@@ -101,6 +101,25 @@ export default class Pubsub {
     return this.subscribe(channel, callback);
   }
 
+  /**
+   * Subscribe the channel once.
+   * This function takes one message off from a pubsub channel,
+   * returning a promise of that message.
+   *
+   * @param {string} channel Channel to listen on
+   * @return {Promise} Promise of next message in this channel
+   */
+  once(channel) {
+    const self = this;
+    return new Promise(function (resolve) {
+      function handler(data) {
+        self.unsubscribe(channel, handler);
+        resolve(data);
+      }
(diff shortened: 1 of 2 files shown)


oursky-travis:
@akiroz, thanks for your PR! By analyzing the history of the files in this pull request, we identified @cheungpat to be a potential reviewer.

cheungpat:
Thanks for your pull request. Can you add a test case?

I think it is useful to a one-off message from a channel and don’t need to bother to subscribe/unsubscribe. I see you are kind enough to add esdoc to the function, but can you elaborate a bit on what the function does? Describe the behavior of the function a bit so that the user of the function will know what to expect.

akiroz:
I've updated the function description, but still not sure how I should setup the test after looking through `test/pubsub.js` for a while....

cheungpat:
Since you are going to call `subscribe` and `unsubscribe` of container, you probably should use sinon to mock these two functions and check if they are called with the required parameters.

Then you should call `ws.onmessage` to simulate a message received from the websocket. You should check the returned promise to see if it resolves to the expected message.

See http://sinonjs.org/releases/v2.1.0/spies/ under “Using a spy to wrap an existing method”.

akiroz:
@cheungpat tests added~

cheungpat:
This is merged. Thanks.

Can you write some documentation on how this works? I think the right location to add doc is here: https://github.com/SkygearIO/guides/blob/master/pubsub/basics/js.md

PR #187, opened 2017-04-11
Fix auth provider not binding when calling `handleAction`

ben181231:
This will cause a crash while auth providier will call
`this[action]` to get corresponding action method


oursky-travis:
@ben181231, thanks for your PR! By analyzing the history of the files in this pull request, we identified @rickmak to be a potential reviewer.

rickmak:
Merge via 8954b0e

Issue #186, opened 2017-04-10
Container saves and get user info with a shared global store


b123400:
Reported from user, all containers are sharing the same store. It is impossible to have 2 containers logged in with different users

https://skygear.slack.com/archives/C1AD5Q51R/p1491823247658724?thread_ts=1491580427.456695&cid=C1AD5Q51R

https://github.com/SkygearIO/skygear-SDK-JS/blob/5527d5251675cd4a0b8258b6215988d6bfbfde4c/lib/container.js#L653

PR #188, opened 2017-04-12
Fix issue with shared user between containers

b123400:
connect https://github.com/SkygearIO/skygear-SDK-JS/issues/186

### DIFF
--- a/lib/cloud/container.js
+++ b/lib/cloud/container.js
@@ -16,7 +16,7 @@
 /* eslint camelcase: 0 */
 const _ = require('lodash');
 
-import Container from '../container';
+import Container, {USER_CHANGED} from '../container';
 
 export default class CloudCodeContainer extends Container {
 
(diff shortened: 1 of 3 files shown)


oursky-travis:
@b123400, thanks for your PR! By analyzing the history of the files in this pull request, we identified @cheungpat to be a potential reviewer.

cheungpat:
@b123400 The test program doesn‘t exit after running the test case. Can you take a look?

cheungpat:
In the long run, I think we should have the following hierarchy. But it is okay as is:

* CommonContainer with all the functions for calling actions and handlers
  * Container for client development
  * CloudContainer for cloud code development, with user session storage applicable in cloud code environment

b123400:
@cheungpat Agree, but a little suggestion on naming.
The word "common" should be avoided, because it only means codes is shared between something, any class with subclasses can be called "common".
I think it should be "ClientContainer", "CloudContainer" and "Container".

Issue #160, opened 2017-01-11
Consider switching to use explicit react related config instead of implicit determine


rickmak:
In `container.js` and `store.js`, there is a `require('react-native');` clause. The cause create `Cannot find module` in some environment, like ionic. It will required extra configuration to set `react-native` to `undefined` when using webpack bundling.

- Skygear SDK Date/Version: 0.21.0
- JS runtime:
- Skygear Server Date/Version: 0.21.0
- [ ] Is this a regression?
- [ ] Attached logs, screenshots

david90:
@rickmak How to workaround at this moment?

rickmak:
It depends on what bundler you are using. For webpack, you can mark react-native as external to make  the bundle works. If you are using generator, we do it for you: https://github.com/SkygearIO/generator-skygear/blob/master/generators/index.js#L114

david90:
@rickmak  Using Ionic (ionic1), when require using browerify, it cannot find module.

calvinchankf:
@rickmak it works on webpack1 but how to make it work on webpack2?

I tried the same as the skygear boilerplate
```
externals: {
    'react-native': undefined,
    'websocket': undefined
  }
```
but it shows WebpackOptionsValidationError when I compile the project
https://drive.google.com/file/d/0B4GsONVca8IOSDNfN01Yc04yWmc/view

Instead, I tried
```
externals: [ 'reactive', 'websocket' ]
```
It compiled successfully but my project doesn't recognize `react` instead
https://drive.google.com/file/d/0B4GsONVca8IOalJVZDFKLV81QWc/view?usp=sharing

Please help

b123400:
I sent a PR https://github.com/SkygearIO/skygear-SDK-JS/pull/206 that works around the issue. It's preventing webpack from bundling react-native, which means it is not following the suggestion solution in this issue. Because:

- Less config is better
- Config is not always available
  - When an angular project is created via the official `angular-cli`, it does not expose any option to config the packaging process, [intentionally](https://github.com/angular/angular-cli/issues/1656#issuecomment-239366723).

Actually after this PR, angular / ionic / webpack does not need any extra config to use skygear, which means a normal `require('skygear')` is already good enough. Maybe we don't need https://github.com/SkygearIO/guides/issues/58 as well.

chpapa:
> Maybe we don't need SkygearIO/guides#58 as well.

We could then focus the issue on a react.js guide / boilerplate...?

[...]

PR #206, opened 2017-04-25
Work around webpack trying to bundle react-native

b123400:
connect https://github.com/SkygearIO/skygear-SDK-JS/issues/160

This PR is not exactly doing what the issue is saying, but should solve the problem.

### DIFF
--- a/lib/container.js
+++ b/lib/container.js
@@ -36,7 +36,12 @@ import {EventHandle} from './util';
 
 var React;
 if (typeof window !== 'undefined') {
-  React = require('react-native');
+  // Use indirect call to prevent webpack from looking for react-native
+  try {
+    React = (0, require)('react-native');
+  } catch (e) {
+    // Ignore
+  }
 }
 
 export const USER_CHANGED = 'userChanged';
--- a/lib/store.js
+++ b/lib/store.js
@@ -472,7 +472,13 @@ class Store {
 /* global window: false */
 if (typeof window !== 'undefined') {
   // env: browser-like
-  var rn = require('react-native');
+  // Use indirect call to prevent webpack from looking for react-native
+  var rn;
+  try {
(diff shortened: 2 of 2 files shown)


oursky-travis:
@b123400, thanks for your PR! By analyzing the history of the files in this pull request, we identified @limouren to be a potential reviewer.

louischan-oursky:
I tried this patch but it seems that it does not work with react native packager.

Here are the steps

1. Build this patch and [commit the build](https://github.com/louischan-oursky/skygear-SDK-JS/commits/require-react-native)
2. Initialize a react native project and configure `skygear` to point to the above commit
3. Run the react native app

It should fail with the following log

```
2017-04-26 19:09:26.808690 MyApp[16449:6446190] 'Failed to get user', { [ReferenceError: Can't find variable: document]
  line: 98986,
  column: 32,
  sourceURL: 'http://192.168.1.127:8081/index.ios.bundle?platform=ios&dev=false&minify=false' }
```

I went through the line and column in the log and discovered that the storage is incorrectly recognized as `CookieStorage` (used in lib/store.js) which makes use of `document` which does not exist in react native.

b123400:
TLDR: I give up working around webpack.

-------

- webpack translates all modules' name into number, then require them by the numbers in run time. Something like this:

      require('a') // becomes require(1);
      require = id => {1:module_a()}[id]()

- React native is using webpack with aliasing plugin, it changes the name of a bundle according to `@providesModule`
  - `react-native` is aliased to `react-native-implementation`
- `require('react-native')` doesn't work because it is aliased
- By working around webpack, the "string to number" translation is missing, and it is still requiring `react-native` after built, so it does not work.
- Tried to `require('react-native-implementation')`, it works. But then got a message saying loading module with string does not work in production environment.
- I give up working around webpack

PR #209, opened 2017-04-28
Separate react-native related logic from main code

b123400:
connect https://github.com/SkygearIO/skygear-SDK-JS/issues/160

Looked at how lerna manages packages and I don't think we should use it. It puts packages into the `packages` folder, and there is no "root" package.

E.g. In react, the root level is a package called [`react-build`](https://github.com/facebook/react/blob/master/package.json), the real `react` is at `packages/react`. If I have to use lerna, I will need to move all files to a subdirectory, and therefore update all build tools and might break something.

Moreover, I realise we already have an existing pattern for solving differences between platforms, which is the `lib/cloud` folder. By asking using to `require('skygear/cloud')` we can do whatever injection we like, so I implemented the react-native part via `require('skygear/react-native')`.

### DIFF
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import store from './store';
+import getStore from './store';
 
 export default class Cache {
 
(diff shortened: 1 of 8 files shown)


oursky-travis:
@b123400, thanks for your PR! By analyzing the history of the files in this pull request, we identified @louischan-oursky to be a potential reviewer.

rickmak:
Merge via 8f56847

Issue #214, opened 2017-05-08
[Cloud Code] S3 signer does not respect url prefix


louischan-oursky:
1. It should respect url prefix but what it generates is plain S3 url.
2. It always generate http url even url prefix is https

PR #215, opened 2017-05-08
Make S3 signer respect url prefix

louischan-oursky:
connect #214

### DIFF
--- a/lib/cloud/asset.js
+++ b/lib/cloud/asset.js
@@ -171,18 +171,9 @@ export class S3Signer extends Signer {
 
     const key = this.assetStoreS3AccessKey;
     const secret = this.assetStoreS3SecretKey;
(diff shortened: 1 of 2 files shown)


oursky-travis:
@louischan-oursky, thanks for your PR! By analyzing the history of the files in this pull request, we identified @rickmak to be a potential reviewer.

louischan-oursky:
If the prefix includes path, then it is not preserved.

e.g. `https://your-cdn.com/` is preserved but `https://your-cdn.com/static/` is not.

Though I don't know whether prefix is supposed to have path.

louischan-oursky:
@rickmak I just found a bug in my original implementation. It does not handle trailing slash properly. I have updated the PR to fix that.

rickmak:
Merge via 8560c9c

PR #217, opened 2017-05-12
Make this repo ready for multiple-packages

b123400:
connect https://github.com/SkygearIO/skygear-SDK-JS/issues/192

This is a part of https://github.com/SkygearIO/skygear-SDK-JS/issues/192
Before implementing it, we have to make this repo multiple-package, such that we can have `skygear-core`, `skygear-forget-password`, etc, in the same repo.In the future we can put more things into the `skygear` package while keeping the core small.

The public API of skygear should remain the same after this PR.

Notice that:

- ESdoc generation only works for `skygear-core`, not other packages. Seems like esdoc.org only supports one `esdoc.json` per repo.
- I have not tried the deploy command, any test server to try...?
- Docker image is only built for `skygear` (not `skygear-core`), is that ok?


oursky-travis:
@b123400, thanks for your PR! By analyzing the history of the files in this pull request, we identified @rickmak to be a potential reviewer.

Makefile
```diff
@@ -23,11 +24,11 @@ test:
 
 .PHONY: clean
 clean:
-	-rm -rf dist
+	-rm -rf packages/*/dist
 
 .PHONY: build
 build:
-	$(DOCKER_RUN) npm pack
+	$(DOCKER_RUN) cd packages/skygear && npm pack
```
cheungpat: This line expands to the following when `WITH_DOCKER=1`
```
docker run --rm -i -v `pwd`:/usr/src/app -w /usr/src/app skygeario/skygear-nodedev cd packages/skygear && npm pack
```

That means `cd` is treated as a command but it is actually a shell builtin function. Wrap it inside `sh -c "cd ..."` like above.

b123400: Oh right, updated

rickmak:
We need to update `example`. Which is auto deploy to `https://sdkjsexample.skygeario.com/static/index.html` by `scripts/deploy-example.sh`

> ESdoc generation only works for skygear-core, not other packages. Seems like esdoc.org only supports one esdoc.json per repo.

 I am fine with this for now.

> I have not tried the deploy command, any test server to try...?

Add a temp command
```
if [ "$TRAVIS_BRANCH" == "travis" ]; then
  npm run deploy
fi
```
at `scripts/deploy.sh` and try it on Traivs?

PS: I try at my local with `npm run deploy` and got `Error: Cannot find module 'skygear-core'`.

rickmak:
Btw, we will need to update this files too: https://github.com/SkygearIO/skygear-server/blob/master/RELEASING.md

b123400:
Oops, forget to update instruction. You need to run `lerna bootstrap` (or `npm bootstrap` if you don't have lerna installed globally) before building, the command install dependencies for all packages, and link them together.

Sorry I used the wrong wording, by "any test server to try...?", I mean is there any s3 bucket I can try to upload? I don't want to deploy to the production s3 bucket because this is not merged to master yet.

> We need to update example. Which is auto deploy to https://sdkjsexample.skygeario.com/static/index.html by scripts/deploy-example.sh

Oh right, updated.

rickmak:
> I don't want to deploy to the production s3 bucket because this is not merged to master yet.

Just use tag and let me to delete afterward. Don't worth the trouble to open an s3 for you.

rickmak:
Where is the update of `scripts/deploy-example.sh`

PR #220, opened 2017-05-17
Add skygear-forgot-password package

b123400:
connect https://github.com/SkygearIO/skygear-SDK-JS/issues/192

I am adding the module via

```
module.exports.forgotPassword = require('skygear-forgot-password');
```

Seems not very good, but where else should I inject...?

### DIFF
--- a/packages/skygear-forgot-password/lib/index.js
+++ b/packages/skygear-forgot-password/lib/index.js
@@ -1,0 +1,14 @@
+import skygear from 'skygear-core';
+
+export const forgetPassword = (email, container = skygear) =>
+  container.lambda('user:forgot-password', {
+    email: email
+  });
+
+export const resetPassword = (userID, code, expireAt, newPassword, container = skygear) =>
+  container.lambda('user:reset-password', {
+    user_id: userID,     /* eslint camelcase: 0 */
+    code: code,          /* eslint camelcase: 0 */
+    expire_at: expireAt, /* eslint camelcase: 0 */
+    new_password: newPassword
+  });
--- a/packages/skygear-forgot-password/lib/index.js
+++ b/packages/skygear-forgot-password/lib/index.js
@@ -1,14 +1,24 @@
 import skygear from 'skygear-core';
 
-export const forgetPassword = (email, container = skygear) =>
-  container.lambda('user:forgot-password', {
+function _forgetPassword(email) {
+  return this.lambda('user:forgot-password', {
     email: email
   });
+}
 
-export const resetPassword = (userID, code, expireAt, newPassword, container = skygear) =>
-  container.lambda('user:reset-password', {
+function _resetPassword(userID, code, expireAt, newPassword) {
+  return this.lambda('user:reset-password', {
     user_id: userID,     /* eslint camelcase: 0 */
     code: code,          /* eslint camelcase: 0 */
     expire_at: expireAt, /* eslint camelcase: 0 */
     new_password: newPassword
   });
+}
+
+export const forgetPassword = _forgetPassword.bind(skygear);
+export const resetPassword = _resetPassword.bind(skygear);
+
+export const injectToContainer = (container = skygear) => {
+  container.constructor.prototype.forgetPassword = _forgetPassword;
+  container.constructor.prototype.resetPassword = _resetPassword;
+};
--- a/gulp/tasks/test.js
+++ b/gulp/tasks/test.js
@@ -28,10 +28,10 @@ gulp.task('pre-test', function () {
     .pipe(istanbul({
       includeUntested: true,
       instrumenter: isparta.Instrumenter
-    }))
-    .pipe(istanbul.hookRequire());
+    }));
   });
-  return merge(streams);
+  return merge(streams)
+    .pipe(istanbul.hookRequire());
 });
 
 gulp.task('test', ['pre-test'], function (cb) {
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "require-dir": "^0.3.0",
     "sinon": "^1.17.2",
     "sinon-chai": "^2.8.0",
+    "superagent": "^3.4",
     "vinyl-buffer": "^1.0.0",
     "vinyl-source-stream": "^1.1.0",
     "watchify": "^3.3.1"
--- a/packages/skygear-forgot-password/package.json
+++ b/packages/skygear-forgot-password/package.json
@@ -1,33 +1,22 @@
 {
-  "name": "skygear",
+  "name": "skygear-forgot-password",
   "version": "0.23.0",
-  "description": "JS SDK of Skygear services",
+  "description": "Forgot password for Skygear JS SDK",
   "homepage": "https://github.com/SkygearIO/skygear-SDK-JS",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/SkygearIO/skygear-SDK-JS.git"
   },
   "author": "rickmak <rick.mak@gmail.com> (https://skygear.io)",
-  "files": [
-    "cloud.js",
-    "bin.js",
-    "asset.js",
-    "react-native.js",
-    "dist"
-  ],
   "main": "dist/index.js",
   "keywords": [
     "SDK",
-    "skygear"
+    "skygear",
+    "forgot password"
   ],
-  "bin": {
-    "skygear-node": "./bin.js",
-    "skygear-asset": "./asset.js"
-  },
   "license": "Apache-2.0",
   "dependencies": {
-    "skygear-core": "^0.23.0",
-    "skygear-forgot-password": "^0.23.0"
+    "skygear-core": "^0.23.0"
   },
   "bugs": {
     "url": "https://github.com/SkygearIO/skygear-SDK-JS/issues"
--- a/packages/skygear/lib/index.js
+++ b/packages/skygear/lib/index.js
@@ -1,1 +1,6 @@
 module.exports = require('skygear-core');
+
+const forgotPassword = require('skygear-forgot-password');
+for (let key in forgotPassword) {
+  module.exports[key] = forgotPassword[key];
+}
--- a/packages/skygear/lib/index.js
+++ b/packages/skygear/lib/index.js
@@ -1,6 +1,6 @@
-module.exports = require('skygear-core');
-
+const container = require('skygear-core');
 const forgotPassword = require('skygear-forgot-password');
-for (let key in forgotPassword) {
-  module.exports[key] = forgotPassword[key];
-}
+
+forgotPassword.injectToContainer(container);
+
+module.exports = container;


oursky-travis:
@b123400, thanks for your PR! By analyzing the history of the files in this pull request, we identified @rickmak to be a potential reviewer.

packages/skygear-forgot-password/lib/index.js
```diff
@@ -0,0 +1,13 @@
+import skygear from 'skygear-core';
+
+export const forgetPassword = (email, container=skygear)=>
```
rickmak: `(email, container=skygear)=>` -> `(email, container=skygear) =>`

rickmak:
In the feature issue. The example code is `skygear.forgotPassword(“chpapa@gmail.com”)` which your way of exporting will works. What is the issue?

cheungpat:
@b123400 this PR needs rebasing

b123400:
Rebased and fixed test cases.

@rickmak `skygear.forgotPassword(...)` actually didn't work. It needs to be

```
skygear.forgotPassword.forgotPassword(...)
skygear.forgotPassword.resetPassword(...)
```

Because I was importing it like this:

```
module.exports.forgotPassword = require('skygear-forgot-password');
```

In order to make the sample code works, I made it so:

```
const forgotPassword = require('skygear-forgot-password');
for (let key in forgotPassword) {
  module.exports[key] = forgotPassword[key];
}
```

But there is still problem. If user creates their own Container, that new container will not have the `forgotPassword` method.
They will need to call the function like this:

```
var anotherContainer = new Container();
anotherContainer.forgotPassword // undefined;
skygear.forgotPassword(..., anotherContainer);
```

Which seems strange.

rickmak:
I want to raise the concern why this issue can pass the features phase if the API and use cases are not clearly defined. @chpapa

chpapa:
> I want to raise the concern why this issue can pass the features phase if the API and use cases are not clearly defined. @chpapa

It seems the API have some minor problems. But what's the problem with the use cases?

BTW, actually, I think it makes most sense to put it under Container since other user related functions are there. Is it possible? @b123400

b123400:
Possible, I can "inject" it into the class, I dont think it is very common though...

cheungpat:
> BTW, actually, I think it makes most sense to put it under Container since other user related functions are there. Is it possible? @b123400

I don’t think it is a good idea to have function injected to the container. This is a bad practice.

chpapa:
> I don’t think it is a good idea to have function injected to the container. This is a bad practice.

Any alternative to making the API consistent to the users? Or what are the disadvantages?

I think the arguments to keep it in the container is strong, think about the API, these are all API inside the container:
* changePassword
* loginWithEmail (and other Login functions)
* signupWithEmail (and other Signup functions)
* getUserByEmail

It would be very natural for developers when they look at API doc/guides, they expect forgotPassword / resetPassword are inside Container.

b123400:
I think as long as the injection is explicit / encapsulated, it should be fine?

Encapsulated:
```
skygear = require('skygear')
skygear.forgetPassword() // works
```

Explicit injection:
```
core = require('skygear-core');
core.forgetPassword() // no method

f = require('skygear-forgot-password');
f.forgetPassword() // works
core.forgetPassword() // still no method

f.injectToContainer(core); // need explicit injection
core.forgetPassword() // works now
```

cheungpat:
> I think as long as the injection is explicit / encapsulated, it should be fine?

seems good

chpapa:
@b123400 sorry another consistency issue, so if we have resetpassword and forgotpassword in container for js sdk, do we have the same for iOS and Android sdk?

b123400:
iOS is fine.

```
[container forgotPasswordWithEmail:@"email@domain.com" completionHandler:^(NSDictionary *dict, NSError *error) {
  ...
}];
```

[Android is not.](https://github.com/SkygearIO/skygear-SDK-Android/pull/129)

```
ForgotPassword f = new ForgotPassword(container);
f.forgotPassword("email@domain.com", handler);
```

Java class is not extendable in that way... not sure what is the best practice in the community.

cheungpat:
I would like to draw @rickmak’s attention to this: https://github.com/SkygearIO/skygear-SDK-JS/pull/220#issuecomment-302375169

rickmak:
> It seems the API have some minor problems. But what's the problem with the use cases?

Just want to ensure we are not expecting the API is actually workable in feature issue. Those are pseudocode or a specification of implementable code?

> It would be very natural for developers when they look at API doc/guides, they expect forgotPassword / resetPassword are inside Container.

Your agenda is developer should not aware that function is provided via a plugin. Which is not align with exiting plugin sdk way of doing thing. i.e. We are expecting the dev aware it and do this explicitly `import skygearChat from 'skygear-chat';`

So from now on we will do the separate of module based on business separation, not by programming organization. Agree? (I assume yes. Just to ensure.)

With this agree, @b123400 suggestion on explicit injection will be fine for me in JS context.

> @b123400 sorry another consistency issue, so if we have resetpassword and forgotpassword in container for js sdk, do we have the same for iOS and Android sdk?

Also, we agree previously each SDK will implement according to the target language convention. We are now agree to rate consistency across platform over language idiom. I think this need discussion to which level our API is fluenent in language but still easy to understand for the developer that is working in multipl platform.

In the case of JAVA. I think we will need skygear and skygear-core separation to make it look like extendable?

chpapa:
> Just want to ensure we are not expecting the API is actually workable in feature issue. Those are pseudocode or a specification of implementable code?

It seems better to be more specific? For forgotPassword it was more like a trial run; Actually I think we will have to adjust based on the time spent and how well plan each features would need to be, so I wanted to say "it depends", but as a general principle we should be able to give meaningful sample code in feature issue so it should be more specific.

> Also, we agree previously each SDK will implement according to the target language convention. We are now agree to rate consistency across platform over language idiom. I think this need discussion to which level our API is fluenent in language but still easy to understand for the developer that is working in multipl platform.

I still think we should implement each SDK according to the target language's convention. However, in this case, we're not talking about the convention of JS / iOS / Android are different; The convention is the same across all 3 platforms -- all have classes and a Container class! But we're putting different features in different classes, that's not a difference due to the platform.

So we're definitely not talking about **raising consistency across platform over language idiom**.



> So from now on we will do the separate of module based on business separation, not by programming organization. Agree? (I assume yes. Just to ensure.)

As posted on Slack channel, I think we are forced to pick one of these options in this situations:

1. Give up the "logical consistency" of SDK plugins map to Server SDKs. So we move forgot password into skygear-core SDK.  (Instead, we should keep the "logical consistency" of all plugins forced to enable in Skygear.io, will be in skygear-core; This is different from *separate of module based on business separation* as you said.

2. Inject functions to container uglily, we need to fix problems like how to make the generated SDK API docs include those injected API too btw. And seems we don't have a solution on Android.

3. Keep the logical consistency of SDK to Server plugins mapping. So we move User into its own plugins and have all users functions called from something like SkygearUser instead of Container.

In an ideal world, it seems 3 is better. But unless we are ready to do a huge restructuring of Skygear-Core, I think 1 make more sense.

Clear? @rickmak @b123400 @cheungpat @ben181231

chpapa:
Of coz if anyone come up with a way we can do 2 well / avoid the huge inconsistency between platforms, 2 would still be more preferred...

chpapa:
After the offline discussion today with @rickmak @cheungpat @ben181231, and looking at the current API design from a layman perspective, I have the following thoughts and suggestions. Let me know if you think these make sense?

1. Source of the problem is now `Container` basically contains ALL features in its method list, including DB, pubsub, push notification, User. (Actually, DB is better since they're under `privateDB` and `publicDB`, but we still got ACL related functions in Container.
    1. So whenever we add features that are directly related with DB, pubsub, push, or users, we will face the dilemma of either we present a non-sense API arrangement to users (like we got `forgotPassword` and `resetPassword` in a different class, while we got `changePassword` in Container), or we have to inject everything into Container.
    1. Even if we can inject everything into Container, Container will become a crazy class and impossible to comprehend some day (it is actually quite difficult to understand now anyway)
    1. My understanding is now `Container` represent a session, so we still want to keep this concept.
1. Based on the above, I would propose we should actually restructure `Container` (yea so we can group this breaking changes with new User Profile), to solve the problem.
    1. I think we should have something like `container.pubsub.on()` / `container.publicDB.setDefaultACL()` and `container.auth.loginWithEmail()` etc, so we group different features into its own class.
        1. Probably the exception are `config`, `configApiKey`, `lambda` 
        1. I'm not so sure about `clearCache` / `makeUploadAssetRequest`, and what is `inferDeviceType` / `makeRequest` / `sendRequestObject`?
        1. Do we want to group `SkygearChat` into `container.chat` too?
    1. We should have something like `skygear-auth` which is included automatically by importing `skygear`, and it will have depends on both `skygear-core` and `forgot-password` and even `sso` plugins. 
        1. That's a trade-off, but I suggest we assume it is temporary. We will eventually refactor all auth features into `skygear-auth` on server side too.
        1. Before we able to refactor the above, I suggest we just make it clear on our documentation [guides](https://docs.skygear.io/guides/advanced/server/) and [skygear-server README.md](https://github.com/SkygearIO/skygear-server/blob/master/README.md) what are the minimal dependency of core features.
    1. I want to stress that I think users who sign up at skygear.io and make Skygear as new-developer-friendly as possible, is much more important than making it easy to do local deployment or advanced users who want to disable/enable each plugin. 

I hope this will settle the discussion so that we can move quickly to the actual design and implementation of the features next week. So please feel free to comment and discuss.

b123400:
> I still think we should implement each SDK according to the target language's convention. However, in this case, we're not talking about the convention of JS / iOS / Android are different; The convention is the same across all 3 platforms -- all have classes and a Container class! But we're putting different features in different classes, that's not a difference due to the platform.
>
> So we're definitely not talking about **raising consistency across platform over language idiom**.

The logic of "They all have classes and a Container class, so their conventions are the same" does not make sense.

- In Objective-C, class is designed to be extensible.
- In Javascript, there is no real class, we pretend to have one. Class extension is not a best practice in the community.
- In Java, class is simply not extensible, unless using Reflection.

I think we all agree the goal of plugin is to "add new functionalities", the problem now is **"How to add it?"**. We can extend Container in Objective-C and JS, but not Java, that's just how the languages work, this is language idiom. If we have to make them consistent, we should not depend on class extension, we should depend on something available in every language.

> Source of the problem is now `Container` basically contains ALL features in its method list, including DB, pubsub, push notification, User.

I think this is not true.

- You think the problem is "We are not separating things well".
- I think the problem is "We are not combining things in a consistent way".

> I think we should have something like `container.pubsub.on()` / `container.publicDB.setDefaultACL()` and `container.auth.loginWithEmail()` etc, so we group different features into its own class.
>
> We should have something like `skygear-auth` which is included automatically by importing `skygear`, and it will have depends on both `skygear-core` and `forgot-password` and even `sso` plugins.

I am getting confused by this. Hope you don't mind clarifying it:

- Can user develop an application with only the `skygear-core` package?
  - e.g. They only need a small set of functionality
- If no, what is the use of `skygear-core`?
- If yes, continue:
- Given `skygear-auth` is depending on `skygear-core`, can I assume `skygear-core` is not (circliarly) depending on `skygear-auth`?
- In Java, 
  - In which package is the type `Container` defined?
  - If `Container` is defined in `skygear-core`, what is the type of `container.auth`, when `skygear-auth` is not installed?
  - If `Container` is not defined in `skygear-core`, how can user develop an application with only the `skygear-core` package?


- In JS, What is the type of
  -  `require('skygear-core')`
  -  `require('skygear-auth')`
  -  `require('skygear')`

> As posted on Slack channel, I think we are forced to pick one of these options in this situations:

I don't agree these are the only choices. Here is my proposal:

- Don't extend class.
- Ask user to include plugin manually.
- When using plugin, users has to explicitly get the plugin out.

## JS

```
// Setting up, only need to call once
var skygear = require('skygear');
var skygearForgotPassword = require('skygear-forgot-password');
skygear.include(skygearForgotPassword);

// Usage
var plugin = container.getPlugin('forgotPassword');
plugin.forgotPassword('email');
```

## iOS

```
// Setting up, only need to call once
[Skygear include:[SkygearForgotPassword class]];

// Usage
SkygearForgotPassword *plugin = [container getPlugin:@"forgotPassword"];
[plugin forgotPassword:@"email"];
```

## Java

```
// Setting up, only need to call once
Skygear.include(SkygearForgotPasswordFactory());

// Usage
SkygearForgotPassword plugin = (SkygearForgotPassword)Skygear.getPlugin("forgotPassword");
plugin.forgotPassword("email");
```

chpapa:
> The logic of "They all have classes and a Container class, so their conventions are the same" does not make sense.

There are actually two meaning of "conventions" here:
1. The convention of Skygear SDK
2. The convention of the programming language.

"They all have a Container class, so their conventions are the same" => the convention here meant 1. You're talking about 2, the convention of the programming language.

On 2, Let's say if we decided to inject functions into Container. I would, of course, agree we shall use whatever language's the best way to do that.

For skygear users, on each SDK, they saw us using `Container convention` on each platform already (put every users related functions in that class), and I believe if we break this convention, would make users have a very hard time to understand the features of the API.

> I think this is not true.
> 
> You think the problem is "We are not separating things well".
> I think the problem is "We are not combining things in a consistent way".

I don't know what is the point of this response? Does "We are not separating things well" and "We are not combining things in a consistent way" have any significant difference? It sounds kinda similar to me, lol

> Can user develop an application with only the skygear-core package?

In my proposal, No it can't. The `skygear-core` package could be named something else, maybe more like `skygear-init`/ just `skygear`? It is just for initializing the SDK.

> If no, what is the use of skygear-core?

I think you're referring to this: "and it will have depends on both skygear-core and forgot-password and even sso plugins." -- actually, I meant depends on both `skygear-server`, which is on server side. (edited to make it more readable)

> In Java,

In Java, I think it could works like some standard DI framework. Without any implementation "injected", `skygear.auth` should be null. However we should assume that the init should got the basic features imported already.

> In JS, What is the type of

Sorry I'm not familiar with modern JS enough to answer this I think, however I believe JS have no problem implement the `container` class suggested? Or what is the intend of your question?

> I don't agree these are the only choices. Here is my proposal:
> 
> Don't extend class.
> Ask user to include plugin manually.
> When using plugin, users has to explicitly get the plugin out.

However this choice doesn't align with giving users a consistent API on client side. We will end up have `changePassword` in Container, and `forgotPassword` in something else. Or `loginWithEmail` vs `loginWithGoogle`.

I think the major difference between my proposal and your proposal is, I wish our API doc is simple enough for users to glace through it quickly and understand what they can do with Skygear (at least for most common things); And they don't have to read through tons of guides to just understand `getPlugin`, what are the plugins available, and how to call them.

cheungpat:
I agree with everything that @b123400 [wrote](https://github.com/SkygearIO/skygear-SDK-JS/pull/220#issuecomment-302802237) and he is spot on for raising those questions.

That being said, it appears the concerns raised and proposals put forward by @b123400 are not satisfactory from @chpapa point of view.

I think what @chpapa wants are these:

* A `Container` class that exposes all API.
* The API have to be discoverable by the app developer, probably by way of looking at class doc, or through IDE autocompletion.
* The API have to be grouped into feature-set so that app developer can find an API by narrowing down to the set of API via the grouping. Grouping is achieved by way of putting API of the same feature-set into a separate class to be exposed by the `Container`.
* The API exposed by all languages SDK have to be the same, meaning the same set of API has to be available within the same grouping and exposed via similar mechanisms.
* The API organization is totally app-developer-driven, so how the plugin code is organized doesn’t matter.

I think @chpapa wants the code to look like this:

> Note: this is not real code, do not copy from these code, these are for illustration purpose only.

```objective-c
// Objective-C
#import <SKYKit/SKYKit.h>

SKYContainer *container = [SKYContainer defaultContainer];
[container.auth loginWithEmail:@"email@example.com" password:@"secret" completion:nil];
[container.auth forgotPasswordWithEmail:@"email@example.com" completion:nil];
[container.pubsub subscribeTo:"channel" handler:nil];
```

```java
// Java
import android.content.Context;
import io.skygear.skygear.Container;
import io.skygear.skygear.Pubsub;

Container container = Container.defaultContainer(Context.getApplicationContext());
container.auth.loginWithEmail("email@example.com", "secret");
container.auth.forgotPassword("email@example.com");
container.pubsub.subscribe("channel", new Pubsub.Handler())
```

```javascript
// JS
var skygear = require('skygear');

skygear.auth.loginWithEmail("email@example.com", "secret");
skygear.auth.forgotPassword("email@example.com");
skygear.pubsub.on("channel", function() {});
```

(I am not even sure this is what @chpapa wants, because while they adhere to the language naming convention, they are not consistent across SDKs.)

---

Here comes the technical part. As a good developer, I want to adhere to as many good conventions and practices as I could practically possible while staying within the framework of API design I stated within the beginning of this comment.

Objective-C is easy because class can be extended and extending class is part of the language convention (albeit debatable). The `SKYKit` cocoapod will include as many sub-specs and dependencies as needed. The sub-spec (such as `SKYKit/ForgotPassword`) will extend the relevant classes to expose the API as illustrated above.

In JS, while it is possible to inject function into existing class, this does not adhere to convention. To achieve the above, the `skygear` package will require as many other sub-packages and dependencies as needed in order to expose all API through the `Container` class in the `skygear` package. If `skygear-forgot-password` is its own package, the `skygear` package will require it and expose the `forgotPassword` function through `container.auth` (`container.auth` is part of `skygear` package).

```javascript
// In skygear package
var fp = require('skygear-forgot-password');

class Container {
  get auth() {
    return this._auth;
  }
}

class ContainerAuth {
  forgotPassword(email) {
    return fp.forgotPassword(email);
  }
}
```

In Java, it is not possible to extend Java class. We could do that in a way similar to Javascript in which the `io.skygear.skygear` package imports as many other sub-packages and dependencies as needed. We can also use Dependency Injection but we need a clear and stable `interface`. Without a clear and stable `interface`, I don’t have an opinion of whether we should do DI.

In other words, for all SDKs, `skygear` is a fat package comprising of all the APIs that the app developer may want to use. Doing so makes the `skygear` package “fat” because it will bring in as many dependencies as needed to exposes all the API we will implement, no matter in core or as a separate plugin.

If an app developer think the `skygear` comprises of too many dependencies, they will need to cherry-pick a collection of sub-packages such as `skygear-pubsub`, `skygear-auth`, `skygear-db` to require/import. Depending on the language, the developer may need to implement their own version of `Container` (because the `Container` in `skygear` package require/import other features that the developer don’t need).

---

I presented this comment in order to bridge the differences. I don’t have a personal opinion in the future direction of the SDK.

b123400:
> If an app developer think the skygear comprises of too many dependencies, they will need to cherry-pick a collection of sub-packages such as skygear-pubsub, skygear-auth, skygear-db to require/import. Depending on the language, the developer may need to implement their own version of Container

I see, it makes sense now. I had an impression that `skygear-core` alone has to be useful enough for developers to use because of the original intent of splitting modules.

------

As long as we make it clear that the responsibility of "making skygear client slimmer" is at the developer's side, I am fine with the direction.

rickmak:
> Note: this is not real code, do not copy from these code, these are for illustration purpose only.

```objective-c
// Objective-C
#import <SKYKit/SKYKit.h>

SKYContainer *container = [SKYContainer defaultContainer];
[container.auth loginWithEmail:@"email@example.com" password:@"secret" completion:nil];
[container.auth forgotPasswordWithEmail:@"email@example.com" completion:nil];
[container.pubsub subscribeTo:"channel" handler:nil];
```

```java
// Java
import android.content.Context;
import io.skygear.skygear.Container;
import io.skygear.skygear.Pubsub;

Container container = Container.defaultContainer(Context.getApplicationContext());
container.auth.loginWithEmail("email@example.com", "secret");
container.auth.forgotPassword("email@example.com");
container.pubsub.subscribe("channel", new Pubsub.Handler())
```

```javascript
// JS
var skygear = require('skygear');

skygear.auth.loginWithEmail("email@example.com", "secret");
skygear.auth.forgotPassword("email@example.com");
skygear.pubsub.on("channel", function() {});
```

This looks good to me. Let do it? Any objection?

cheungpat:
> This looks good to me. Let do it? Any objection?

@chpapa I went ahead and created a new issue https://github.com/SkygearIO/features/issues/70 for the changes proposed earlier. If it doesn’t fit your mental model, we can always cancel that and see how we can come up with something better.

chpapa:
> @chpapa I went ahead and created a new issue SkygearIO/features#70 for the changes proposed earlier. If it doesn’t fit your mental model, we can always cancel that and see how we can come up with something better.

Thanks. I just commented there. There is one consistency issue would like to see what you guys think.

Steven-Chan:
see #240

PR #223, opened 2017-05-18
Fix unable to use SkygearResponse returned from handler

cheungpat:


packages/skygear-core/lib/cloud/transport/common.js
```diff
@@ -104,27 +104,130 @@ class SkygearRequest {
...
+   * be returned.
+   * 
+   * @param result - SkygearResponse or response body
+   * @return {!SkygearResponse} a SkygearResponse
+   */
+  static wrap(result) {
+    if (result instanceof SkygearResponse) {
+      return result;
+    } else if (result.body !== undefined) {
+      return new SkygearResponse(result);
```
b123400: I cannot find this logic in the original implementation (If `body` exists in the first argument, the argument should be treated as the option for `SkygearResponse`).

Wouldn't this make the behaviour strange?

`return {"foo": "bar"}` and `return {"body": "hi", "foo": "bar"}` will mean different things, but it is hard to tell.

cheungpat: > Wouldn't this make the behaviour strange?

Yes.

I need a way to recognize that the returned object from a handler is of type SkygearResponse. As far as I know, in JS, the way a JS is imported will affect the `result instanceof SkygearReponse` check, meaning that I am not able to accurately determine if the object in question is a SkygearResponse. A fallback to that is to check if the object has a `body` field.

(The case that returning a `SkygearResponse` from a handler but the transport not recognising that the object is of type `SkygearResponse` usually happens on the cloud. This is because the cloud runs `npm install skygear` and `npm install`, so it is possible to end up with two copies of `SkygearResponse` class.)

I am open to suggestion on how best to accomplish this.

b123400: How about something like this?

```
class SkygearResponse {
  constructor() {
    this._isSkygearResponse = true;
  }

  static isInstance(obj) {
    return (obj instanceof SkygearResponse) || obj._isSkygearResponse;
  }
}

if (SkygearResponse.isInstance(result)) {
  ...
}
```

cheungpat: Much better.

Issue #231, opened 2017-06-02
Update SDK API Grouping


Steven-Chan:
https://github.com/SkygearIO/features/pull/71

- [ ] Update Container api with the following grouping
- [ ] Update Test
- [ ] Update Example

Grouping:
```
Container:

  state:
    url
    apiKey
    timeoutOptions
    auth(AuthContainer)
    relation(RelationContainer)
    db(DatabaseContainer)
    pubsub(PubsubContainer)
    push(PushContainer)
    chat(ChatContainer)
    analytics(AnalyticsContainer)

  api:
    config
    configApiKey
    lambda
    sendRequestObject
    makeRequest


AuthContainer:

  state:
    accessToken
    user(User)

  api:
    signupWithUsername
    signupWithEmail
    signupWithUsernameAndProfile
    signupWithEmailAndProfile
    signupAnonymously
    loginWithUsername
    loginWithEmail
    loginWithProvider
    logout
    changePassword
    forgotPassword
    // following api would be affected by https://github.com/SkygearIO/features/pull/68
    saveUser
    whoami
    getUsersByEmail
    getUsersByUsername
    discoverUserByEmails
    discoverUserByUsernames


RelationContainer:

  api:
    queryFriend
    queryFollower
    queryFollowing
    addRelation
    removeRelation


DatabaseContainer:

  state:
    public(PublicDatabase)
    private(PrivateDatabase)
    cacheResponse(Boolean)

  api:
    makeUploadAssetRequest


PublicDatabase:

  state:
    cacheStore

  api:
    setAdminRole
    setDefaultRole
    getDefaultACL
    setDefaultACL
    setRecordCreateAccess
    setRecordDefaultAccess
    getRecordByID
    save
    query
    del
    clearCache
    fetchSubscriptionWithID
    saveSubscription
    deleteSubscriptionWithID


PrivateDatabase:

  state:
    cacheStore

  api:
    getRecordByID
    save
    query
    del
    clearCache
    fetchSubscriptionWithID
    saveSubscription
    deleteSubscriptionWithID


PubsubContainer:

  state:
    channel
    internalChannel
    autoPubsub(Boolean)

  api:
    on
    off


PushContainer:

  api:
    sendPushNotification


ChatContainer:


AnalyticsContainer:

```

Mobile:
```
PushContainer:

  state:
    deviceToken
    deviceID

  api:
    registerDevice
    unregisterDevice
```

Example:
```
// Call API

[...]

PR #234, opened 2017-06-06
Update container API grouping

Steven-Chan:
connect #231


packages/skygear-core/lib/container.js
```diff
@@ -673,77 +242,121 @@ export default class Container {
...
+     * @property {number} [timeoutOptions.response=60000] - maximum time to
+     * wait for an response (in milliseconds)
+     *
+     * @see http://visionmedia.github.io/superagent/#timeouts
+     */
+    this.timeoutOptions = {
+      response: 60000
+    };
+  }
+
```
rickmak: Please add esdocs annotation to `Container`. At least to make the doc navigtable, and we can fill in the detail laters.

packages/skygear-core/lib/database.js
```diff
@@ -33,6 +33,10 @@ export default class Database {
     this._cacheResponse = true;
   }
 
+  get Record() {
+    return this.container.Record;
+  }
```
rickmak: I am in double the Record are accessible from boath BaseContainer and database container. Semm we should choose one. In this case, I will choose BaseContainer.

Steven-Chan: Well, what if this is for internal use? Like `User` and `publicDB` in `AuthContainer`. (I think these are two different cases, `User` can actually be imported directly, but `publicDB` cannot.)

rickmak: Mark it as private by annotation or prefix with an underscore?

Steven-Chan:
@rickmak Updated

I've also made some extra changes on the skygear.relation on 333676d2753f6c74b096012603eef43d093966a1

example/index.html
```diff
@@ -78,7 +78,7 @@
     <hr>
     <p><button onclick="logout();">Logout</button></p>
   </div>
-  <script src="../dist/bundle.js"></script>
+  <script src="../packages/skygear-core/dist/bundle.js"></script>
```
cheungpat: good catch

Issue #250, opened 2017-07-18
New user profile design: Update skygear.auth api


Steven-Chan:
feature: https://github.com/SkygearIO/features/issues/48
server issue: https://github.com/SkygearIO/skygear-server/issues/393

- [x] add `signup(authData, password, [profile])`
- [x] add `login(authData, password)`
- [x] update current `signWithXXX` and `loginWithXXX` to call new server api
- [x] return `User` record in signup, login and whoami response
- [x] remove APIs that are calling server `user:xxx`
  - [x] remove `saveUser`
  - [x] remove `getUserByXXX`
- [x] remove `User` object (the original user object returned in signup/login api), the followings are needed to replace the api provided by the original `User` object
  - [ ] add `add/get/deleteRoles` api to `skygear.auth`, work with user id and role object
  - [x] make `ACL` and `Relation` object consumes `User` record instead of `User` object

Steven-Chan:
getRoles api would require https://github.com/SkygearIO/skygear-server/issues/430

PR #256, opened 2017-07-25
New signup login, remove user object

Steven-Chan:
connect #250


Steven-Chan:
@rickmak ready for review

rickmak:
I think by semantic, `assignUserRole` and `revokeUserRole` should not belong to publicDB? It is more like `auth` things.

My categorisation:
- DB is managing all things that have JIT and save to db without `_` prefix.
- auth is managing user access level related meta.

rickmak:
I am expecting another PR to address the Role related API call regrouping.

PR #257, opened 2017-07-25
Fix exception on server fails to bind address and return meaningful message

rickmak:
On server error, the callback only provide `err` and the socket is already
closed. We should only just log out the error for user to inspection.


ben181231:
Picked as 161d184

Issue #267, opened 2017-07-28
Upload asset without also saving a record


cheungpat:
This is a feature request: In CMS, asset is uploaded when the user select a file from the file picker. The user may decide to save the record with the new asset or the user may abort and not save anything at all.

This function is previously exposed by `skygear.makeUploadAssetRequest` which is now marked internal in version 1.0.0. To support the CMS use case, I suggest we create a `skygear.publicDatabase.uploadAsset` to upload an asset without also trigger a record save.

rickmak:
I am wondering should the upload Asset belong to publicDatabase. Feel like we should have a `skygaer.asset.upload`.


Because publicDatabase is only record related, putting asset to this grouping is no good. In addition, the publicity of a asset is control by setting rather than by publicDatabase/privateDatabase. So it worth to make a new group `asset`. also it give hit to developer he can use skygear as asset store just as s3.

cheungpat:
The `uploadAsset` function is best contained in `DatabaseContainer` (which is not `PublicDatabase` nor `PrivateDatabase`) but it this container is not exposed by Container. 

I could create a `DatabaseContainer` by `new skygear.DatabaseContainer(container)` and then execute the `uploadAsset` function from there, if we decide this is not a common use case.

PR #268, opened 2017-07-28
Add api uploadAsset to skygear database

Steven-Chan:
connect #267

### DIFF
--- a/packages/skygear-core/lib/container.js
+++ b/packages/skygear-core/lib/container.js
@@ -222,6 +222,41 @@ export class BaseContainer {
     return ErrorCodes;
   }
 
+  /**
+   * @type {AuthContainer}
+   */
+  get AuthContainer() {
+    return AuthContainer;
+  }
+
+  /**
+   * @type {RelationContainer}
+   */
+  get RelationContainer() {
+    return RelationContainer;
+  }
+
+  /**
+   * @type {DatabaseContainer}
+   */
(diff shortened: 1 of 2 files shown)


rickmak:
Merge via 873cd73

cheungpat:
For me to access the function `uploadAsset`, I need access to `DatabaseContainer`, which isn’t exposed as far as I know.

Steven-Chan:
@cheungpat what about this?
```
import {DatabaseContainer} from 'skygear/database';

const dbContainer = new DatabaseContainer(skygear);
dbContainer.uploadAsset(...);
```

cheungpat:
> @cheungpat what about this?

is this intended usage?

cheungpat:
@Steven-Chan I need it in browser-land.

i.e. all i have is `<script src="https://cdn.example.com/skygear.min.js"></script>`

Steven-Chan:
@cheungpat 
If we decide not to make upload asset a separated functionality, the undocumented way to do it now is `skygear._db.uploadAsset`.

Otherwise, I would prefer what @rickmak said in #267.

Steven-Chan:
@cheungpat updated

cheungpat:
@rickmak

Steven-Chan:
follow up pr: https://github.com/SkygearIO/skygear-SDK-JS/pull/270

Issue #278, opened 2017-08-03
Include SDK version number to allow feedback from developers


Steven-Chan:
```js
import skygear from 'skygear';

console.log(skygear.version); // v1.0.0 || v1.1.0-alpha.0 || ...
```

cheungpat:
actually kind of good

PR #279, opened 2017-08-04
Inject version string to skygear container with preprocess

Steven-Chan:
connect #278 

example:
```js
import skygear from 'skygear';
// or with window.skygear

console.log(skygear.VERSION);
```


rickmak:
Good, we need --version for skygear-node too.

rickmak:
Merge via 270067e

Issue #284, opened 2017-08-08
Browserify error when deploying 


cheungpat:
In [travis](https://travis-ci.org/SkygearIO/skygear-SDK-JS/jobs/262108228):
```
[05:59:09] Browserify Error { Error: Cannot find module 'skygear-core' from '/home/travis/build/SkygearIO/skygear-SDK-JS/packages/skygear/lib'
    at /home/travis/build/SkygearIO/skygear-SDK-JS/node_modules/browser-resolve/node_modules/resolve/lib/async.js:46:17
    at process (/home/travis/build/SkygearIO/skygear-SDK-JS/node_modules/browser-resolve/node_modules/resolve/lib/async.js:173:43)
    at ondir (/home/travis/build/SkygearIO/skygear-SDK-JS/node_modules/browser-resolve/node_modules/resolve/lib/async.js:188:17)
    at load (/home/travis/build/SkygearIO/skygear-SDK-JS/node_modules/browser-resolve/node_modules/resolve/lib/async.js:69:43)
    at onex (/home/travis/build/SkygearIO/skygear-SDK-JS/node_modules/browser-resolve/node_modules/resolve/lib/async.js:92:31)
    at /home/travis/build/SkygearIO/skygear-SDK-JS/node_modules/browser-resolve/node_modules/resolve/lib/async.js:22:47
    at FSReqWrap.oncomplete (fs.js:123:15)
```

PR #286, opened 2017-08-08
Run lerna bootstrap in postinstall stage

cheungpat:
This fixes the problem when npm install runs the prepublish stage,
the dependencies used by skygear-core is not already installed.

connects #284


rickmak:
Merge via 81a5f0a

PR #288, opened 2017-08-08
Add node 8 to travis-CI env matrix

rickmak:


cheungpat:
Looks like skygear-SDK-JS can build on node 8 and I think we should remove node 4 tests and just use node 8 for deployment (edit `deploy:` section)

rickmak:
@cheungpat updated.

Issue #273, opened 2017-08-01
Cloud op `options.authRequired` not working


Steven-Chan:
cloud code:
```
const skygearCloud = require('skygear/cloud');

function includeme(skygearCloud) {
  skygearCloud.op('greeting', function(param) {
    return {
      'content': 'Hello, ' + param.name,
    };
  }, {
    authRequired: true,
  });
}

module.exports = {
  includeme
};
```

curl cmd:
```
curl -X POST http://localhost:3000/greeting
```

It is expected to be rejected by the server because no api key is set. But it returns `{"result":{"content":"Hello, undefined"}}`.

PR #293, opened 2017-08-11
Change op and handler option auth_required to key_required

Steven-Chan:
connect #273


Steven-Chan:
@rickmak updated

rickmak:
Close via 7eebf37

Issue #228, opened 2017-05-31
Pub/Sub can only publish JS truthy values


akiroz:
Pub/Sub can only publish JS truthy values

- Skygear SDK Date/Version: 0.23.0

#### Expected Results
EITHER pubsub accept any serializable JS value
OR pubsub message requires a single Object/Array as per the JSON spec

#### Actual Results
`skygear.pubsub.publish` will throw an exception if the message body is JS falsey.
i.e. `42`, `"hello"`, `[]` and `{}` works but `0`, `""`, and `null` does not.

Steven-Chan:
@rickmak I suggest restricting the pubsub message to JSON object/dictionary only to make it consistent across all SDK.

PR #294, opened 2017-08-11
Fix pubsub falsy message, and install missing esdoc plugins

Steven-Chan:
connect #228


cheungpat:
Separate this PR into two.

Issue #113, opened 2016-11-09
Cannot read property 'toJSON' of undefined


lampercy:
```js
const Foo = skygear.Record.extend('foo');
const foo = new Foo();
foo.c = undefined;
skygear.publicDB.save(foo)
>> Uncaught (in promise) TypeError: Cannot read property 'toJSON' of undefined
```
Should undefined be casted to null, or warn them skygear does not support undefined?

rickmak:
We should provide a better error message.

PR #296, opened 2017-08-11
Return meaningful error for undefined record key

Steven-Chan:
connect #113


rickmak:
Other looks good.

Steven-Chan:
@rickmak updated

rickmak:
Merge via c24af27

PR #297, opened 2017-08-11
Fix pubsub falsy message

Steven-Chan:
connect #228


cheungpat:
I couldn’t decide if this is the correct resolution for refs #228, giving this to @rickmak

Issue #307, opened 2017-09-05
Provide Container.push.sendToUser


rickmak:
JS version of this: https://github.com/SkygearIO/py-skygear/blob/master/skygear/action.py

- Skygear SDK Date/Version:
- JS runtime:
- Skygear Server Date/Version:
- [ ] Is this a regression?
- [ ] Attached logs, screenshots

carmenlau:
Need to support both client and js cloud code

cheungpat:
To match feature parity, I will also implement push to device.

PR #315, opened 2017-09-29
Add sendToUser and sendToDevice

cheungpat:
connects #307


cheungpat:
doc update https://github.com/SkygearIO/guides/pull/129

PR #322, opened 2017-10-26
Update toJSON handling for array like object

Steven-Chan:
Mapping keys of the object instead of mapping the object directly.
The map implemention would give integer key for array-like object, while
skygear record does not need to recognize array-like object.

connect #321


rickmak:
Please add test case for this kind of bugs.

Steven-Chan:
@rickmak test case added

rickmak:
Merge via 85a28d3

Issue #305, opened 2017-08-30
[OAuth] API for integrating basic OAuth login flow


carmenlau:
modified on #204 #205

All function are under `skygear.auth`

- `loginOAuthProviderWithPopup(providerID, options)` and `loginOAuthProviderWithRedirect(providerID, options)`
  - Create or login a new skygear user, associated with the provider
  - `providerID` - A string that identify the login provider
  - `options`
    - scope - Override scope if set in portal
    - redirectUri - The URL that skygear will redirect after authorization flow.
  - This function returns a skygear user, and an access token of the service.

```js
  skygear.auth.loginOAuthProviderWithPopup('com.facebook', {
    scope: []
  }).then(function(skygearUser) {

  }).catch(function(error) {

  });

  skygear.auth.loginOAuthProviderWithRedirect('com.facebook', {
    scope: [],
    redirectUri: 'https://app.example.com/signup'
  });

  skygear.auth.getLoginRedirectResult().then(function(skygearUser) {

  }).catch(function(error) {

  });
```

PR #334, opened 2017-11-06
Support loginOAuthProviderWithPopup flow

carmenlau:
connect #305 

For running the sso example html, require #336


rickmak:
I will review it at #337

PR #233, opened 2017-06-06
Support parameterized paths

tatgean:
**Add support to paths like this:**
`/downloads/:platform/:version`

**In skygear js cloud code:**
`skygearCloud.handler('downloads/:platform/:version', ......)`

**Request:**
`/downloads/osx/1.0.0`

**In the handler function:**
`req.params.platform === 'osx'`
`req.params.version === '1.0.0'`

PR #344, opened 2017-11-14
Support parameterized paths

rickmak:
```
skygearCloud.handler('downloads/{platform}/{version}', (req) => {
  // Take request to /downloads/osx/1.1.2?format=zip as an example
  console.log(req.params.platform); // got `osx`
  console.log(req.params.version); // got `1.1.2`
  console.log(req.params.build); // got undefined
});
```

Superset of #233 
connects #344

### DIFF
--- a/packages/skygear-core/lib/cloud/registry.js
+++ b/packages/skygear-core/lib/cloud/registry.js
@@ -144,7 +144,7 @@ export class Registry {
 
   getHandler(path, method) {
     for (let k in this.handlers) {
-      if (this._matchHandler(k, path)) {
+      if (Registry._matchHandler(k, path)) {
         let func = this.handlers[k][method];
         func.handlerName = k;
         return func;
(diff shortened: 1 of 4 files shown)


cheungpat:
Not sure if I follow.

I used the snippet in the PR to try out this PR, the handler is registered at plugin init time:

app_1     | time="2017-12-01T06:59:02Z" level=info msg="Get response from init" err="<nil>" logger=plugin out="{\"op\":[{\"name\":\"greeting\",\"key_required\":true,\"user_required\":false}],\"event\":[{\"name\":\"init\"}],\"handler\":[{\"name\":\"downloads/{platform}/{version}\",\"methods\":[\"GET\",\"POST\",\"PUT\"]},{\"name\":\"static/\",\"methods\":[\"GET\",\"POST\",\"PUT\"],\"user_required\":false}],\"hook\":[],\"timer\":[],\"provider\":[]}" plugin="&{1 0xc42018ea80 map[]}"

But curl'ing has no effect in the console log (according to the snippet, the plugin process should print the platform and version string).

```
$ curl http://localhost:3000/downloads/1/2
{"result":{"status":"OK"}}
```

Looking at the init playload, I’d expect the plugin to register `downloads` instead of `downloads/{platform}/{version}`, because the latter one will cause skygear server to look for this string literally.

curl'ing with this url does match the registered handler:

```
$ curl http://localhost:3000/downloads/\\{platform\\}/\\{version\\}
{"error":{"name":"UnexpectedError","code":10000,"message":"TypeError: Cannot read property 'platform' of undefined"}}
```

Probably missing something in handler registration code? @rickmak

Issue #342, opened 2017-11-12
Why container.pubsub.reconfigure() is required to establish pubsub connection?


cheungpat:
A user reported that the container.pubsub.reconfigure function is required to be called in order to establish the pubsub connection. This issue aims to confirm that the issue is reproducible and check if any resolution is necessary.

- Skygear SDK Date/Version:
- JS runtime:
- Skygear Server Date/Version:
- [ ] Is this a regression?
- [ ] Attached logs, screenshots

#### Steps to reproduce

1.
2.
3.

cheungpat:
I couldn’t see why calling `reconfigure()` is needed as `container.config()` already call the same function internally. Any idea @rickmak?

PR #346, opened 2017-11-15
Fix auth _getUser error handling

Steven-Chan:
This has caused error in skygear.config if skygear-user is not a correct
user, making the pubsub connect not called.

connect #342


rickmak:
Merge via fc9c55b

PR #347, opened 2017-11-17
Re-export SkygearError as skygear.Error

rickmak:
The lib user should not import from either `lib` nor `dist`. Since lib
user should not know the internal structure, we maintains the usage to
be `import skygear from 'skygear';` and use the SkygearError class as
Skygear.Error.


carmenlau:
merged via 2c33b41

PR #353, opened 2017-11-30
Don't set the X-Skygear-Access-Token on null

### LINKED ISSUE #351: Receive `AccessTokenNotAccepted` when query public read records in node env without logged in user
Consider the following query
```js
const CategoryRecord = skygear.Record.extend('category');

const query = new skygear.Query(CategoryRecord);
query.addAscending('order');
skygear.publicDB.query(query)
  .then(r => console.log(r))
  .catch(e => console.log(e));
```
If I have a unlogged-in user, when run it in browser, the query works fine. However, when run it in node server, will receive `AccessTokenNotAccepted`

- Skygear SDK Date/Version: 0.24.2
- JS runtime: node
- Skygear Server Date/Version: 0.24.0
- [ ] Is this a regression?
- [ ] Attached logs, screenshots

#### Expected Results
Can query public read records

#### Actual Results
Received `AccessTokenNotAccepted` exception

cychiuae:
Inspired by #352 to try to fix #351 for version under `1.0.0`

### DIFF
--- a/packages/skygear-core/lib/container.js
+++ b/packages/skygear-core/lib/container.js
@@ -458,8 +458,10 @@ export default class Container {
     let req = this.request
       .post(this.url + _action)
       .set('X-Skygear-API-Key', this.apiKey)
-      .set('X-Skygear-Access-Token', this.accessToken)
       .set('Accept', 'application/json');
+    if (this.accessToken) {
+      req = req.set('X-Skygear-Access-Token', this.accessToken)
+    }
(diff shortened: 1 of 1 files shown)


rickmak:
Merge via 688fdfd

rickmak:
@cychiuae do you think releasing a 0.24.3 will help you? If yes, We can do ao patch release to npm

cychiuae:
@rickmak If so, it will be perfect. 
No need to worry. I have done a workaround to this problem

PR #354, opened 2017-12-03
Fix cannot config skygear container for chat in react native

carmenlau:
connect SkygearIO/chat-SDK-JS#78

- Removed `ReactNativeContainer` and reuse the same container for react-native
- Removed `_getAccessToken` and `_getDeviceID` in auth and push container constructor, since they will cause errors in react-native as the store is not ready before calling config. After this change, access token and device id will only be ready after calling config.

Both setups are supported
```
import skygear from 'skygear/react-native';
import skygearChat from 'skygear-chat';

skygear.config({...}).then(() => {
  skygearChat.createDirectConversation(...);
});
```

```
import skygear from 'skygear';
import skygearChat from 'skygear-chat';

skygear.config({
  ...,
  platform: 'react-native'
}).then(() => {
  skygearChat.createDirectConversation(...);  // I would get api key not configured error here
});
```

### DIFF
--- a/packages/skygear-core/lib/container.js
+++ b/packages/skygear-core/lib/container.js
@@ -84,12 +84,6 @@ export class BaseContainer {
      * @private
      */
     this.ee = ee({});
-
-    /**
-     * Platform support default and react-native
-     * @private
-     */
-    this.platform = 'default';
   }
 
   /**
@@ -124,9 +118,6 @@ export class BaseContainer {
     }
     if (options.endPoint) {
       this.endPoint = options.endPoint;
-    }
-    if (options.platform) {
-      this.platform = options.platform;
     }
     return Promise.resolve(this);
   }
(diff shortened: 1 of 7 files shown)


packages/skygear-core/lib/container.js
```diff
@@ -510,6 +528,30 @@ export default class Container extends BaseContainer {
...
+  /**
+   * @private
+   *
+   * Configure container based on the platform.
+   *
+   * This will be called in config function when the platform option change.
+   * Or you can call it after change the platform.
+   * Currently only react native is supported.
+   */
+  configPlatform() {
```
rickmak: `configPlatform` -> `_configPlatform`

I am not sure why you design this function without taking a parameter. I think taking a parameter is better. i.e. `_configPlatform(platform)`. Also, move the `changedPlatform` check to this function?

packages/skygear-core/lib/container.js
```diff
@@ -493,10 +504,17 @@ export default class Container extends BaseContainer {
    * @param {Object} options - configuration options of the skygear container
    * @param {String} options.apiKey - api key
    * @param {String} options.endPoint - end point
+   * @param {String} options.platform - default or react-native
```
rickmak: `default` means....? Web or Node?

Issue #325, opened 2017-10-30
Only one handler is registered if two with the same endpoint but different request method is defined


cheungpat:
Looking at the source code in `registry.js`, the JS cloud code runtime also suffers the same bug as it exists in skygeario/py-skygear#164

- Skygear SDK Date/Version:
- JS runtime:
- Skygear Server Date/Version:
- [ ] Is this a regression?
- [ ] Attached logs, screenshots

#### Steps to reproduce

1.
2.
3.

PR #356, opened 2017-12-04
Fix register handler with different HTTP method

cheungpat:
This fixes the bug when registering multiple handler function
of the same path but with different HTTP method.

connects #325


carmenlau:
Merged via 7f01b80

PR #365, opened 2018-01-08
Fix Record construct with falsy attrs

### LINKED ISSUE #361: react-native: Warning shown when container configure if container is not logged in with a user
- Skygear SDK Date/Version: v1.2.0
- JS runtime: React-Native
- Skygear Server Date/Version:
- [ ] Is this a regression?
- [ ] Attached logs, screenshots

#### Expected Results

No warning.

#### Actual Results

Error message:

```
Failed to get user TypeError: Cannot read property 'id' of null
```

#### Steps to reproduce

1. first time open app, or logout and re-open app

inDream:
Fix #361 

When code compiled, the default attribute only check for `undefined` but not `null`.
```js
// original
class Record {
  constructor(recordType, attrs = defaultAttrs) {
  }
}
// compiled
function Record(recordType) {
    var attrs = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : defaultAttrs;
}
```

According [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/default_parameters):  
Default function parameters allow formal parameters to be initialized with default values if **no value or undefined** is passed.

When auth init with no user, `null` is passed into Record constructor and triggered error.
https://github.com/SkygearIO/skygear-SDK-JS/blob/master/packages/skygear-core/lib/auth.js#L383
```js
return this.container.store.getItem('skygear-user').then((userJSON) => {
  let attrs = JSON.parse(userJSON); // attrs = null
  this._user = new this._User(attrs);
  return this._user;
})
```

### DIFF
--- a/packages/skygear-core/lib/record.js
+++ b/packages/skygear-core/lib/record.js
@@ -78,10 +78,13 @@ const _metaKey = _.map(_metaAttrs, function (obj) {
  */
 export default class Record {
 
-  constructor(recordType, attrs = defaultAttrs) {
+  constructor(recordType, attrs) {
     if (!Record.validType(recordType)) {
       throw new Error(
         'RecordType is not valid. Please start with alphanumeric string.');
+    }
+    if (!attrs) {
+      attrs = _.assign({}, defaultAttrs);
     }
     this._recordType = recordType;
     // Favouring `id`, since `id` will always contains type information if
--- a/packages/skygear-core/test/record.js
+++ b/packages/skygear-core/test/record.js
@@ -34,6 +34,23 @@ describe('Record', function () {
     );
   });
 
+  it('handle falsy attrs', function () {
+    let r = new Record('user', null);
+    let tuple = r.id.split("/");
+    expect(v4Spec.test(tuple[1])).to.be.true();
+    expect(tuple[0]).to.equal('user');
+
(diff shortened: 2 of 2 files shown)


Steven-Chan:
Thanks @inDream. However, test case should be added for this update.

@rickmak 

What's your thought on this fix? @inDream and I discussed briefly, and decided that Record should also handle `null`. But the current fix check for falsy value, not just `null`. 

I think this fix is somewhere in the middle.

For the current direction, the Record constructor should be more powerful to reject all unexpected value, not just rejecting falsy value. 

Another direction is to keep the good old Record constructor, and handle the `null` value when getting value from local store.

IMO, either of the direction is ok.

cheungpat:
> Another direction is to keep the good old Record constructor, and handle the null value when getting value from local store.

I think this is a better option. I also asked @rickmak and he thinks this is how the behaviour should be.

I agree with @Steven-Chan that this pull request need test cases. It would be great if @inDream could continue working on this pull request to add test cases testing fasly values as parameter to record constructor.

This pull request will not fix #361 though, please @inDream makes another pull request for checking if the value returned from local storage is `null` before creating a user record in `_getUser()` function.

Let me know if you have questions. 👍

PR #373, opened 2018-01-17
Try to fix RN Android got INVALID_STATE_ERR exception

cychiuae:


rickmak:
We are muting the error here, which is not good.
I think the `INVALID_STATE_ERR` is because we don't clean up callbacks.

Closing and expecting an other PR.

PR #374, opened 2018-01-17
Fix pubsub websocket state maybe inconsitent

cychiuae:
A proper fix for bug mentioned in #373


rickmak:
Landed at v0 branch at 6cde08d; Landed at current master at f4ad1d4

PR #379, opened 2018-01-24
Fix signup profile serialization

Steven-Chan:
connect skygeario/skygear-server#524


carmenlau:
Breaks into https://github.com/SkygearIO/skygear-SDK-JS/pull/381 and https://github.com/SkygearIO/skygear-SDK-JS/pull/382

PR #384, opened 2018-02-04
Use docker multi-stage build

cheungpat:
This also changes the user that the node user is running in. Previously
the image default user is root, while the new default user is node.
Skygear packages is now installed under home directory.


cheungpat:
@carmenlau

cheungpat:
@carmenlau updated

Issue #391, opened 2018-02-15
includeme should wait for Promise finish


inDream:
When include modules configuration before lambda / hook can register e.g. database ORM, Promise would not handled inside includeme.

- Skygear SDK Date/Version: 1.3.1
- JS runtime: Node.js 8.9.4
- Skygear Server Date/Version: 1.3.3
- [ ] Is this a regression?
- [ ] Attached logs, screenshots

#### Expected Results
Container could get DB instance.
#### Actual Results
DB would not get instance.

#### Steps to reproduce
```js
const { getContainer } = require('skygear/cloud');
const container = getContainer();
module.exports = {
  includeme() {
    return DataStore.getInstance().then((db) => {
      container.db = db;
    });
  },
};
```

Ref: #158

inDream:
Closed in dd1ec76

PR #398, opened 2018-03-06
Includeme handle promise

inDream:
Connects #391

### DIFF
--- a/packages/skygear-core/lib/bin.js
+++ b/packages/skygear-core/lib/bin.js
@@ -43,12 +43,12 @@ if (cmd === '--settings') {
   process.exit();
 }
 
-settings.loadModules.forEach(function (moduleName) {
-  configModule(moduleName);
-});
+const loadModulesPromise = Promise.all(
+  settings.loadModules.map((moduleName) => configModule(moduleName)),
+);
 
 const codePath = path.join(process.cwd(), cmd);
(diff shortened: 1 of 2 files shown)


packages/skygear-core/lib/cloud/index.js
```diff
@@ -421,7 +421,7 @@ export function configModule(moduleName, options) {
   const { includeme } = require(moduleName);
   if (includeme !== undefined) {
     const settings = {};
-    includeme(module.exports, settings);
+    return new Promise(resolve => resolve(includeme(module.exports, settings)));
```
rickmak: We are not going to promisfy the return config module. We just want the transport start after the promise resolved if it return a promise. so we should just return the value here: `return resolve(includeme(module.exports, settings));`

packages/skygear-core/lib/bin.js
```diff
@@ -68,4 +68,8 @@ if (settings.http.enabled) {
   throw new Error('Currently, only http transport is supported.');
 }
 
-transport.start();
+loadModulesPromise
```
rickmak: Let group up all the promise logic here.

rickmak:
Please follow up at https://github.com/SkygearIO/skygear-SDK-JS/pull/400

PR #400, opened 2018-03-07
Support configMoudle and main entry point return promise

rickmak:
The transport will only start after the promise is resolved


packages/skygear-core/lib/bin.js
```diff
@@ -68,4 +65,12 @@ if (settings.http.enabled) {
   throw new Error('Currently, only http transport is supported.');
 }
 
-transport.start();
+Promise.all(loadModule)
```
inDream: `loadModule` -> `loadModules`

cheungpat:
s/configMoudle/configModule/ (in the commit line)

rickmak:
@cheungpat Updated. Thx.

Issue #401, opened 2018-03-09
Current user not set to null when logout request fail


inDream:
- Skygear SDK Date/Version: 1.4.0
- JS runtime: React Native
- Skygear Server Date/Version: 1.4.0
- [ ] Is this a regression?
- [ ] Attached logs, screenshots

#### Expected Results
`skygear.auth.currentUser === null`

#### Actual Results
`skygear.auth.currentUser === prevUser`

#### Steps to reproduce
1. Login
2. Call `skygear.auth.logout()` without network

Ref: https://github.com/SkygearIO/skygear-SDK-JS/blob/8969e7d19b50710750c28ea98682ceb096b88e1c/packages/skygear-core/lib/auth.js#L201-L211

PR #402, opened 2018-03-09
Handle offline logout clear user info

inDream:
Connects #401


carmenlau:
@inDream Thanks so much for the PR. Discussed with the team and they think your fix looks good, could you help fix the test case and then we can merge. Thanks!

PR #403, opened 2018-03-14
Add reset password api

Steven-Chan:


carmenlau:
@Steven-Chan Sorry for late response, please help add example. Let's have a new page for admin api?

Issue #406, opened 2018-03-21
Rename disable user api with admin prefix


carmenlau:
(Describe the issue here)

- Skygear SDK Date/Version:
- JS runtime:
- Skygear Server Date/Version:
- [ ] Is this a regression?
- [ ] Attached logs, screenshots

#### Steps to reproduce

1.
2.
3.

PR #407, opened 2018-03-28
Add admin prefix to disable/enable user functions

cheungpat:
connects #406

### DIFF
--- a/packages/skygear-core/lib/auth.js
+++ b/packages/skygear-core/lib/auth.js
@@ -348,10 +348,12 @@ export class AuthContainer {
   /**
    * Enable user account of a user.
    *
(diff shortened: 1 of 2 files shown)


packages/skygear-core/lib/asset.js
```diff
@@ -58,8 +58,6 @@ export default class Asset {
       // do nothing
     } else if (base64) {
       file = base64StringtoBlob(base64);
-    } else {
-      throw new Error('Either file or url should present');
```
Steven-Chan: Why having this change?

cheungpat: lol

Issue #392, opened 2018-02-26
Cannot upload base64 image from cloud code


carmenlau:
**Should check both cloud and react-native when handling this issue**

- Skygear SDK Date/Version: 1.3.1
- JS runtime: cloud
- Skygear Server Date/Version:
- [ ] Is this a regression?
- [ ] Attached logs, screenshots

#### Expected Results
Should be able add asset attachment to note

#### Actual Results
Error occur during upload image for base 64 image

#### Sample cloud code

```
const skygear = require('skygear');
const skygearCloud = require('skygear/cloud');

const container = new skygearCloud.CloudCodeContainer();
container.apiKey = 'api_key';
container.endPoint = skygearCloud.settings.skygearEndpoint + '/';

skygearCloud.handler('/hello/', function (req) {
    const picture = new skygear.Asset({
        name: 'hello.png',            // filename of your asset
[...]

Issue #251, opened 2017-07-18
Unable upload Asset record from Node


yipsang:
(Describe the issue here)

- Skygear SDK Date/Version: 1.0
- JS runtime: node v8.1.3
- Skygear Server Date/Version:
- [x] Is this a regression?
- [ ] Attached logs, screenshots

```
var skygear = require('skygear');

var createDummySCN = function() {
	const SCN = skygear.Record.extend('scn');

	const scnAsset = new skygear.Asset({
		name: 'dummy.scn',
		base64: '4pyTIMOgIGxhIG1vZGU=',
		contentType: 'application/octet-stream'
	});
		
	const scn = new SCN({
		name: 'dummy',
		attachment: scnAsset 
	});

	return scn;
}

console.log("Creating dummy scn and texture record ...");

skygear.config({
	'endPoint': 'xxx', // trailing slash is required
	'apiKey': 'xxx',
}).then(() => {
	return skygear.auth.loginWithUsername('admin', 'secret');
}).then(() => {
	console.log("Saving scn record to db ...");

	const scn = createDummySCN();
	return skygear.publicDB.save(scn);
}).then((record) => {
	console.log(record);
	console.log("Record attachment: " +record.attachment.url);
}).catch((err) => {
	console.error('Error in creating scn: ' + err.message);
})
```

rickmak:
The asset can be uploaded, but the asset model is created. 

The https://github.com/SkygearIO/skygear-SDK-JS/blob/7c516a0/packages/skygear-core/lib/database.js#L117 is expecting a file object. The `asset.file` is a `Blob` file which is not confront with the interface that superagent is expecting. Thus will rasie exception on `on` and `pause` method not defined.

chpapa:
@rickmak I don't understand what do you meant in your last comment. What do you meant by "The asset can be uploaded, but the asset model is created."?

chpapa:
And shall we close this issue? Please advices.

rickmak:
> The asset can be uploaded, but the asset model is created.

Should be -> The asset can't be uploaded, the asset model is created.

i.e. as asset model in Postgres is created but pointing to a file does not exist.

It need to be address.

Side note: browser run time work fine.

carmenlau:
Should be fixed in https://github.com/SkygearIO/skygear-SDK-JS/pull/412

PR #412, opened 2018-04-20
Fix cannot upload base64 image in cloud

carmenlau:
connect #392 
connect #251


packages/skygear-core/lib/cloud/asset.js
```diff
@@ -228,3 +229,14 @@ export function getSigner() {
...
   return sharedSigner;
 }
+
+export class Asset extends BaseAsset {
+  constructor(attrs) {
+    super(attrs);
+
+    let base64 = attrs.base64;
+    if (base64) {
+      this.file = Buffer.from(base64, 'base64');
```
cheungpat: Is it possible to move this into `/lib/asset.js` (the non-cloud part of the asset)? I hope it is possible to detect if `Buffer` is available and automatically use `Buffer` if it is available.

Issue #420, opened 2018-05-16
 [Structured logging, Cloud] Review all logs to use new format log functions


carmenlau:
Use function in #419

cheungpat:
I have problem using winston to print error message.

Before using winston,


After using winston,
```
plugin_1  | error: Catching unexpected error: TypeError: Cannot read property 'user_id' of undefined {"meta":{"exception":"{}"},"tag":"plugin","logger":"plugin","exception":{},"timestamp":"2018-05-23T10:33:42.848Z"}
```

Issue #419, opened 2018-05-16
 [Structured logging, Cloud] Log formatting to support JSON and plain text log for different deployment


carmenlau:
https://github.com/SkygearIO/features/blob/master/features/79-structured-logging/implementation.md#js-log-function

PR #424, opened 2018-05-24
Add logger for structural logging

cheungpat:
connects #419
connects #420


packages/skygear-core/lib/cloud/logging.js
```diff
@@ -0,0 +1,116 @@
...
+      ...fields
+    } = normalizeLogFields(rec);
+    /* eslint-enable no-unused-vars */
+    if (!_.isEmpty(fields)) {
+      console.log(`[${level}] ${tag}: ${msg}, ${JSON.stringify(fields)}`);
+    } else {
+      console.log(`[${level}] ${tag}: ${msg}`);
+    }
+    if (error) {
+      console.error(error.stack);
```
Steven-Chan: I think it should either
- the log level to `console` should follow the incoming log level from the logger, or 
- all log to `console.log`. 

And I prefer the first one.
And same should apply to `JSONStream`.

carmenlau:
@Steven-Chan Please check if the update looks good for you? :)

Issue #376, opened 2018-01-18
Support query combination "and"


josephchow102:
Currently, queries are combined by `AND` by default, and can combine two queries with `OR` through `.or(q1, q2)`. However, since there is no `.and(q1, q2)`, some use cases are not covered or become over-complicated.

1. (q1 OR q2) AND q3:

```javascript
let q1 = new Query(Table).equalsTo('field1', 'value1')
let q2 = new Query(Table).equalsTo('field1', 'value2')
let oredQuery = Query.or(q1, q2)
let q3 = oredQuery.equalsTo('field2', 'value3')
```

2. q1 AND (q2 OR q3): not supported

3. (q1 OR q2) AND (q3 OR q4) AND q5: not supported

4. (q1 OR q2) AND (q3 OR q4) AND ... AND qN: not supported

carmenlau:
+1  for CMS references filter support. We need combine the `or` references predicates with the original query

Steven-Chan:
no change needed in skygear-sdk-ios since `NSPredicate` support `NSCompoundPredicate` natively. 

```
NSPredicate *predicate = [NSPredicate predicateWithFormat:@"name = %@ && age >= %d", @"Peter", 12];
```

is the same as 

```
NSPredicate *p1 = [NSPredicate predicateWithFormat:@"name = %@", @"Peter"];
NSPredicate *p2 = [NSPredicate predicateWithFormat:@"age >= %d", 12];
NSPredicate *predicate = [NSCompoundPredicate andPredicateWithSubpredicates:@[p1, p2]];
```

carmenlau:
@Steven-Chan I just merged the android SDK update, would like to know is there any update for JS SDK? Cannot find the relates PR :P

PR #427, opened 2018-06-11
Implement Query.and

Steven-Chan:
connect #376

### DIFF
--- a/packages/skygear-core/lib/query.js
+++ b/packages/skygear-core/lib/query.js
@@ -504,6 +504,12 @@ export default class Query {
     return this;
   }
 
+  _andQuery(queries) {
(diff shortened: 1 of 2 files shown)


packages/skygear-core/lib/query.js
```diff
@@ -659,25 +665,60 @@ export default class Query {
   }
 
   /**
-   * Returns a disjunctive query from queries.
+   * Check if the queries have the same record type
    * @param {Query} queries - Queries
    */
 
-  static or(...queries) {
+  static isSameRecordType(...queries) {
```
carmenlau: It would be better to make this function private to avoid confusing the user?

Steven-Chan:
@carmenlau updated

Issue #429, opened 2018-06-11
Allow user to access uploadAsset function from db


carmenlau:
Currently JS SDK only allow user to upload asset by saving a record with asset. We have the uploadAsset in `DatabaseContainer` which is private and not accessible. 

Allow user to access `uploadAsset` is useful if user want to upload asset before saving record.

#### Expected Results

Can upload asset through `skygear.publicDB.uploadAsset`

#### Actual Results

`uploadAsset` function is under DatabaseContainer and not accessible

PR #430, opened 2018-06-14
Allow user to access uploadAsset function from db

cheungpat:
connects #429

### DIFF
--- a/packages/skygear-core/lib/database.js
+++ b/packages/skygear-core/lib/database.js
@@ -142,7 +142,7 @@ export class Database {
    */
   _presaveSingleValue(value) {
     if (isAsset(value) && value.file) {
-      return makeUploadAssetRequest(this.container, value);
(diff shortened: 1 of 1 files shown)


packages/skygear-core/lib/database.js
```diff
@@ -142,7 +142,7 @@ export class Database {
    */
   _presaveSingleValue(value) {
     if (isAsset(value) && value.file) {
-      return makeUploadAssetRequest(this.container, value);
+      return this.uploadAsset(this.container, value);
```
carmenlau: `uploadAsset` only accept asset, `this.container` is not needed

carmenlau: `uploadAsset` only accept asset, `this.container` is not needed

cheungpat:
@carmenlau fixed

carmenlau:
Merged via 2972049

PR #439, opened 2018-07-04
Throw asset s3 error

Steven-Chan:
connect oursky/skygear-content-manager#269


carmenlau:
- Need update to return skygear error
- Check the s3 payload with entity to large and return skygear error with new error code file size exceeds

packages/skygear-core/lib/database.js
```diff
@@ -557,7 +559,16 @@ function makeUploadAssetRequest(container, asset) {
 
       _request.end((err) => {
         if (err) {
-          reject(err);
+          _parseS3XmlError(err)
```
cheungpat: this should change to async/await

carmenlau:
@Steven-Chan Please update

PR #441, opened 2018-07-05
Check window closure more frequently

cheungpat:
This should fix the test case where the test timed out waiting for the
window to close.


carmenlau:
Merged via 5395440

PR #445, opened 2018-07-13
Use new record ID representation

ben181231:
connect #443

PR #447, opened 2018-07-22
Update npm, fix audit and update build system for node 10

cheungpat:


cheungpat:
if this is merged, remind me to create a new pull request for the v1 branch

ben181231:
Seems there is some permission issues and the `postinstall` script does not run after `make vendor`.

See if [this commit](https://github.com/SkygearIO/skygear-SDK-JS/pull/445/commits/5ad4ae3) fix the issue.

By the way, shall we add `npm run audit` in the test job?

cheungpat:
> Seems there is some permission issues and the postinstall script does not run after make vendor. See if this commit fix the issue.

I don’t have the same problem...

cheungpat:
@ben181231 see if this works better

PR #461, opened 2018-09-07
Export getSigner

louischan-oursky:
Currently we have to write 

```
import { getSigner } from "skygear-core/dist/cloud/asset";
```

to use signer.

it is much better to write

```
import { getSigner } from "skygear/cloud";
```

because importing from `dist` assumes the underlying file structure which may change at any time.


carmenlau:
Merged to master via c1a1790

carmenlau:
Merged to v1-branch via 6f17c70

Issue #462, opened 2018-09-15
Fail to bundle in react native


cychiuae:
Cannot bundle my js code since introduction of `xml2js` in skygear-core . 

- Skygear SDK Date/Version: 1.6.3
- JS runtime: react-native
- Skygear Server Date/Version: N/A
- [x] Is this a regression?
- [x] Attached logs, screenshots

![simulator screen shot - iphone 8 - 2018-09-16 at 03 12 35](https://user-images.githubusercontent.com/10144357/45589780-9274e980-b95e-11e8-9a86-d7d6522562f8.png)

#### Expected Results
Can bundle with skygear sdk

#### Actual Results
Fail to bundle

#### Steps to reproduce

1. `react-native init HelloWorld`
2. `yarn add skygear`
3. import skygear in index.js to ensure skygear will be bundled

Steven-Chan:
@cychiuae have you tried this? https://github.com/Leonidas-from-XIV/node-xml2js/issues/363#issuecomment-318935604

Steven-Chan:
I just found a xml2js fork for react-native https://github.com/connected-io/react-native-xml2js.

Steven-Chan:
@cychiuae could you help me verify if https://github.com/jindw/xmldom works on react-native? It says it works on node and browser.

louischan-oursky:
@Steven-Chan FYI https://www.npmjs.com/package/fast-xml-parser

PR #464, opened 2018-10-08
Fix react-native build problem

Steven-Chan:
connect #462


louischan-oursky:
@tensiuyan @Steven-Chan 

I tested this locally with React Native 0.57 and it now bundles without problems!

cychiuae:
When will it be deployed?

tensiuyan:
@louischan-oursky: thanks! @cychiuae: next week!

Issue #469, opened 2018-11-09
pubsub URL does not respect configured endpoint


louischan-oursky:
(Describe the issue here)

- Skygear SDK Date/Version: any
- JS runtime: any
- Skygear Server Date/Version: any
- [ ] Is this a regression?
- [ ] Attached logs, screenshots

#### Expected Results
Suppose the endpoint is `http://example.com/skygear`
The pubsub URL should be `http://example.com/skygear/pubsub` and `http://example.com/skygear/_pubsub`

#### Actual Results
The path is ignored.

louischan-oursky:
Let me add more details

This is how pubsub URL is constructed https://github.com/SkygearIO/skygear-SDK-JS/blob/v1.6.3/packages/skygear-core/lib/pubsub.js#L82

This is how other URLs is constructed https://github.com/SkygearIO/skygear-SDK-JS/blob/v1.6.3/packages/skygear-core/lib/container.js#L189

pubsub URL is constructed by parsing SKYGEAR_ENDPOINT and ignoring path.

other URLs are constructed by respecting SKYGEAR_ENDPOINT and appending the given path.

PR #470, opened 2018-11-19
Fix pubsub url path

Steven-Chan:
Only host of container endpoint was used in pubsub url, but the path should also be added.

connect #469


carmenlau:
Merged to v1-branch via 7c5f510

carmenlau:
Merged to master via 40ae2aa

PR #473, opened 2018-11-29
Rewrite examples for next

IniZio:


example2/index.html
```diff
@@ -0,0 +1,383 @@
...
+            </div>
+            <div class="row info-row">
+              <div class="col">Access token</div>
+              <div id="info-token" class="col-8" style="word-break: break-all;"></div>
+            </div>
+          </div>
+
+          <script>
+            function refreshUserInfo() {
+            $('#info-token')[0].innerText = skygear.auth.accessToken;
```
carmenlau: Since we use the example as a quick start of using the JS SDK. In order to keep the example simple and easy to read, please avoid using 3rd party library in the SDK example functions. For example, if we only use jquery for finding the dom elements, maybe we just use document.getElementById and give element a clear and easy to understand id. Easy to read is important, as user will read the example to know how to use the api

I think it is ok to use library in the case that help make the ui better, for example `scrollspy`. If possible, we can group those layout setup functions in a js file. So we can reuse it in different pages and user doesn't need read it.

example2/index.html
```diff
@@ -0,0 +1,383 @@
...
+              <div id="info-email" class="col-8"></div>
+            </div>
+            <div class="row info-row">
+              <div class="col">Access token</div>
+              <div id="info-token" class="col-8" style="word-break: break-all;"></div>
+            </div>
+          </div>
+
+          <script>
+            function refreshUserInfo() {
```
carmenlau: Please check all the functions indentation.

example2/index.html
```diff
@@ -0,0 +1,383 @@
...
+              $('#info-username')[0].innerText = '';
+              $('#info-email')[0].innerText = '';
+            }
+          }
+
+          skygear.config({
+            'endPoint': '{{ SKYGEAR_ENDPOINT }}',
+            'apiKey': '{{ SKYGEAR_API_KEY }}'
+          }).then(function(container) {
+            console.log()
```
carmenlau: Please remove not used `console.log`

example2/index.html
```diff
@@ -0,0 +1,383 @@
...
+
+            $('#info-endpoint')[0].innerText = container.endPoint;
+            $('#info-api-key')[0].innerText = container.apiKey;
+            refreshUserInfo()
+          }, function(err) {
+            console.log(err);
```
carmenlau: We can use `console.error` here

example2/index.html
```diff
@@ -0,0 +1,383 @@
...
+                <label for="username" class="sr-only">Username</label>
+                <input type="username" class="form-control" id="username" placeholder="Username" required>
+              </div>
+              <button type="submit" class="btn btn-primary mb-2">Discover</button>
+            </form>
+
+            <textarea id="terminal" class="form-control" cols="20" rows="6" readonly></textarea>
+
+            <script>
+              function discoverUserByUsernames() {
```
carmenlau: It is great that we have a clear function name for each demo! 

Maybe it will look even better if we group them all in a single script block and put them at the end of the body. It would be easier for user to have a quick preview of all the api, and easier for them to find what they want. It would be great if we can have comments before the functions too!

example2/index.html
```diff
@@ -0,0 +1,383 @@
...
+
+  <script>
+    $(function() {
+      var navSelector = '#toc';
+      var $myNav = $(navSelector);
+      Toc.init($myNav);
+      $('body').scrollspy({
+        target: navSelector
+      });
+    });
```
carmenlau: Please group all layout setup functions in js file for reuse.

gulp/tasks/example.js
```diff
@@ -3,7 +3,7 @@ var serve = require('gulp-serve');
 var connectInject = require('connect-inject');
 
 gulp.task('example', serve({
-  root: ['./example', './packages/skygear/dist'],
+  root: ['./example2', './packages/skygear/dist'],
```
carmenlau: Please update the folder to `./exmaple/next/`, the new example is for the next version. And I will rename the PR title.

IniZio:
Since record and auth functions are of separate concerns, maybe i separate this PR into two https://github.com/IniZio/skygear-SDK-JS/tree/next-example-auth and https://github.com/IniZio/skygear-SDK-JS/tree/next-example-record?

tensiuyan:
1. Please give more paddings between two blocks.
paddings

2. Please remove index.html and all v1 examples

3. Wrong btn margin/padding (verify.html)
btn

4. Since admin.html requires Master Key, maybe we can add an info box that guides people to input their Master Key to the example
screenshot 2019-01-29 at 4 56 56 pm

IniZio:
RIP my first pr T_T

louischan-oursky:
Sorry for that. I didn't close it myself. It was closed by GitHub when I force-pushed to the `next` branch. Anyway, your contribution is highly appreciated!

Issue #483, opened 2019-01-23
[BUG] Retry register device token will remove topic ID


cychiuae:
If the deviceID is deleted remotely, by apns feedback. When SDK tries to retry registering device token to server, the topic ID will be removed.

That is because in
https://github.com/SkygearIO/skygear-SDK-JS/blob/c800e44c62964c232b7edc82672a42e7bf9982d3/packages/skygear-core/lib/push.js#L89
We don't pass topicID to retry method

- Skygear SDK Date/Version: 1.6.4
- JS runtime: react-native
- Skygear Server Date/Version: 1.6.x
- [ ] Is this a regression?
- [ ] Attached logs, screenshots

#### Expected Results
Device record can be properly set

#### Actual Results
Topic id is removed from device record.

#### Steps to reproduce

1. Register a device token
2. Delete the record in DB
3. Register the device token again

PR #484, opened 2019-01-23
Fix topic removed when retry registering device

cychiuae:
connects #483


carmenlau:
Merged to master via 78dd47f6
Merged to v1-branch via 701a29f

Issue #490, opened 2019-03-12
Remove `UserRecord` wrapper entirely


kaojohnny:
In PR #489, it introduces `UserRecord` while removing `Record` class because auth gear still wraps user object in a `Record` object. 

In this issue, `UserRecord` should revamp to `User` class, and remove useless `Record` related properties.

Issue #477, opened 2018-12-17
Remove invalid APIs from the JS SDK for next


tensiuyan:
Removed:

- cloud
- asset
- pubsub
- push
- ACL
- database
- query and relation
- reference
- record
- Sequence
- UnknownValue
- Geolocation
- adminDisableUser
- adminEnableUser
- adminResetPassword
- assignUserRole
- revokeUserRole
- setDefaultRole
- setAdminRole

PR #493, opened 2019-03-14
next: Remove user record wrapper

kaojohnny:
connects #477
connects #490
requires https://github.com/SkygearIO/skygear-SDK-JS/pull/489
requires https://github.com/SkygearIO/skygear-server/pull/909


packages/skygear-core/lib/user.js
```diff
@@ -0,0 +1,172 @@
...
+    this.update(attrs);
+  }
+
+  /**
+   * Gets all keys of attributes of the records. Skygear reserved keys, that is
+   * underscore prefixed keys, are excluded.
+   *
+   * @type {String[]}
+   */
+  get attributeKeys() {
```
Steven-Chan: The user object put all dynamic fields inside the `metadata`, the object itself is static, so this function can be removed.

Same for the function `update`, which can be simpler I think.

kaojohnny: @Steven-Chan Updated.

PR #500, opened 2019-04-04
next: Update for login ids

kaojohnny:
Require #498 
connects #497


Steven-Chan:
LGTM, i will wait for SkygearIO/skygear-server#922

Issue #507, opened 2019-08-20
Refactor


louischan-oursky:
- [ ] Move `oauthAuthorizationURL` back to core

```typescript
type Options = RedirectOptions | PopupOptions

interface RedirectOptions {
  action: "login" | "link";
  uxMode: "web_redirect";
  callbackURL: string;
}

interface PopupOptions {
  action: "login" | "link";
  uxMode: "web_popup";
}
```

- [ ] Change abstract method `fetch` to property `transport?: typeof fetch`.
- [ ] Introduce method `fetch` on APIClient and Container
  - skygear-specific fetch
  - prepend endpoint automatically
  - accept path (& query) only, throws error on host found
  - injects headers.

PR #509, opened 2019-08-22
Refactor api client

carmenlau:
connect #507 

Node and react-native are not tested, will test node version when integrating sdk to skycli


carmenlau:
Updated:) @louischan-oursky

PR #510, opened 2019-08-23
Refactor ContainerStorage to accept objects instead of json

carmenlau:


packages/skygear-core/src/storage.ts
```diff
@@ -1,13 +1,114 @@
...
+}
+
+function keyOAuthRedirectAction(name: string): string {
+  return `${name}_oauthRedirectAction`;
+}
+
 /**
  * @public
  */
 export class ContainerStorage {
```
louischan-oursky: We better introduce an interface `IContainerStorage` (not sure about this name, we do not have convention to name interface with prefix `I`) and let `Container` depend on this interface instead of this class. This allows us to plug a compatible storage in skycli without having to extend `ContainerStorage`.

louischan-oursky:
Thanks!

Issue #514, opened 2019-09-04
Update session handling


kiootic:
- Handle refresh tokens automatically
- Do not expose access token in API
- Add session APIs
- Set user agent header
- Collect extra info

PR #515, opened 2019-09-05
Update for new session mechanism

kiootic:
ref #514

### DIFF
--- a/packages/skygear-core/src/client.ts
+++ b/packages/skygear-core/src/client.ts
@@ -58,6 +58,7 @@ export abstract class BaseAPIClient {
   accessToken: string | null;
   fetchFunction?: typeof fetch;
   requestClass?: typeof Request;
+  refreshTokenFunction?: () => Promise<boolean>;
 
   constructor(options: {
     apiKey: string;
@@ -79,10 +80,20 @@ export abstract class BaseAPIClient {
     return headers;
   }
 
-  async fetch(input: string, init?: RequestInit): Promise<Response> {
+  async fetch(
+    input: string,
+    init?: RequestInit,
+    options: { autoRefreshToken?: boolean } = {}
+  ): Promise<Response> {
+    if (this.fetchFunction == null) {
+      throw new Error("missing fetchFunction in api client");
+    }
+
     if (this.requestClass == null) {
       throw new Error("missing requestClass in api client");
     }
+
+    const { autoRefreshToken = !!this.refreshTokenFunction } = options || {};
 
     if (typeof input !== "string") {
       throw new Error("only string path is allowed for fetch input");
@@ -96,19 +107,38 @@ export abstract class BaseAPIClient {
       request.headers.set(key, headers[key]);
     }
 
-    if (this.fetchFunction == null) {
-      throw new Error("missing fetchFunction in api client");
+    let response = await this.fetchFunction(request.clone());
+    if (response.status === 401 && autoRefreshToken) {
+      if (!this.refreshTokenFunction) {
+        throw new Error("missing refreshTokenFunction in api client");
+      }
+
+      const tokenRefreshed = await this.refreshTokenFunction();
+      if (tokenRefreshed) {
+        const retryRequest = request.clone();
+        // use latest access token
+        const headers = this.prepareHeaders();
+        for (const key of Object.keys(headers)) {
+          retryRequest.headers.set(key, headers[key]);
+        }
+
+        response = await this.fetchFunction(retryRequest);
+      }
     }
 
-    return this.fetchFunction(request);
+    return response;
   }
 
   protected async request(
     method: "GET" | "POST" | "DELETE",
     path: string,
-    options: { json?: JSONObject; query?: [string, string][] } = {}
+    options: {
+      json?: JSONObject;
+      query?: [string, string][];
+      autoRefreshToken?: boolean;
+    } = {}
   ): Promise<any> {
-    const { json, query } = options;
+    const { json, query, autoRefreshToken } = options;
     let p = path;
     if (query != null) {
       p += encodeQuery(query);
@@ -119,13 +149,17 @@ export abstract class BaseAPIClient {
       headers["content-type"] = "application/json";
     }
 
-    const response = await this.fetch(p, {
-      method,
-      headers,
-      mode: "cors",
-      credentials: "include",
-      body: json && JSON.stringify(json),
-    });
+    const response = await this.fetch(
+      p,
+      {
+        method,
+        headers,
+        mode: "cors",
+        credentials: "include",
+        body: json && JSON.stringify(json),
+      },
+      { autoRefreshToken }
+    );
     const jsonBody = await response.json();
 
     if (jsonBody["result"]) {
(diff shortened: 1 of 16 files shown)


packages/skygear-core/src/container.ts
```diff
@@ -23,11 +23,6 @@ export class AuthContainer<T extends BaseAPIClient> {
     this.currentIdentity = null;
   }
 
-  // @ts-ignore
-  get accessToken(): string | null {
-    return this.parent.apiClient.accessToken;
-  }
```
louischan-oursky: Also hide `accessToken` on `apiClient`

packages/skygear-core/src/client.ts
```diff
@@ -79,11 +80,21 @@ export abstract class BaseAPIClient {
     return headers;
   }
 
-  async fetch(input: string, init?: RequestInit): Promise<Response> {
+  async fetch(
+    input: string,
+    init?: RequestInit,
+    options: { autoRefreshToken?: boolean } = {}
```
louischan-oursky: Having required argument after optional argument is weird. If we make `init` required, the function signature deviates from the original `fetch` function signature.

packages/skygear-core/src/container.ts
```diff
@@ -131,6 +132,37 @@ export class AuthContainer<T extends BaseAPIClient> {
     this.parent.apiClient.accessToken = null;
   }
 
+  /**
+   * @internal
+   */
+  async _refreshAccessToken(): Promise<boolean> {
+    await this.parent.storage.delAccessToken(this.parent.name);
```
louischan-oursky: Why do we need to delete it first?

packages/skygear-core/src/client.ts
```diff
@@ -82,6 +83,12 @@ export abstract class BaseAPIClient {
     if (this.userAgent !== undefined) {
       headers["user-agent"] = this.userAgent;
     }
+    if (this.getExtraSessionInfo) {
+      const extraSessionInfo = await this.getExtraSessionInfo();
+      if (extraSessionInfo) {
+        headers["x-skygear-extra-info"] = JSON.stringify(extraSessionInfo);
```
louischan-oursky: Do we need to encode the json string with base64 or something else?

packages/skygear-core/src/container.ts
```diff
@@ -27,6 +36,35 @@ export class AuthContainer<T extends BaseAPIClient> {
     this.currentSessionID = null;
   }
 
+  async saveExtraSessionInfoOptions() {
```
louischan-oursky: Why do we need to save this?

packages/skygear-core/src/client.ts
kiootic: `options` is not required?

packages/skygear-core/src/container.ts
kiootic: The access token is no longer valid at this point, so just delete it from storage.

packages/skygear-core/src/client.ts
kiootic: JSON string is safe to be used as HTTP header value.

packages/skygear-core/src/container.ts
kiootic: As mentioned in design, the developer can set the device name once, and the SDK will use the same device name on every app launch.

packages/skygear-core/src/client.ts
louischan-oursky: `init?` is optional

kiootic: I mean both `init` and `options` parameters are already optional.

louischan-oursky: So they are ambiguous.

packages/skygear-core/src/client.ts
louischan-oursky: I guess it is true only when string literal is in unicode code escape.

packages/skygear-core/src/client.ts
kiootic: I don't think it's ambiguous: consumer can specify `input`, `input, init`, or `input, init, options`. Most developer would not change the default options so the usage remains same as `fetch`.

packages/skygear-core/src/client.ts
kiootic: It seems non-ASCII encoded characters in header value is undefined. I'll change to encode the JSON in base64.

packages/skygear-core/src/client.ts
louischan-oursky: I mean the developer cannot call `input, options`

```javascript
function foo(a, b, c = "defaultC") {
  console.log("a", a);
  console.log("b", b);
  console.log("c", c);
}

console.log(`foo("a")`);
foo("a");
console.log();

console.log(`foo("a", "b")`);
foo("a", "b");
console.log();

console.log(`foo("a", "c")`);
foo("a", "c");
console.log();

console.log(`foo("a", "b", "c")`);
foo("a", "b", "c");
console.log();
```

Output
```
foo("a")
a a
b undefined
c defaultC

foo("a", "b")
a a
b b
c defaultC

foo("a", "c")
a a
b c
c defaultC

foo("a", "b", "c")
a a
b b
c c


```

kiootic: I don't think this is a problem. For a majority of developers, `options` need not to be specified. If options need to be changed, requiring developers to use `fetch("URL", undefined, { autoRefreshToken: false })` does not seems unreasonable.

packages/skygear-core/src/container.ts
louischan-oursky: In case the refresh operation failed due to network error, the access token was deleted and the same flow cannot be retried because the client SDK will return error early complaining that access token is null.

kiootic: > the client SDK will return error early complaining that access token is null.

Could you point me to where the client SDK checks the access token is null? The client SDK should not do any check on access token at client-side, since it may stored in cookies.

In your case, the network error would be returned to consumer. If consumer retry the operation, the SDK would not send the access token to server. After receiving a 401 error, the SDK would attempt to refresh the token again.

packages/skygear-core/src/client.ts
```diff
@@ -89,7 +89,7 @@ export abstract class BaseAPIClient {
     if (this.getExtraSessionInfo) {
       const extraSessionInfo = await this.getExtraSessionInfo();
       if (extraSessionInfo) {
-        headers["x-skygear-extra-info"] = JSON.stringify(extraSessionInfo);
+        headers["x-skygear-extra-info"] = btoa(JSON.stringify(extraSessionInfo));
```
louischan-oursky: First, nodejs does not have `btoa`

Second, it is unclear that what encoding the javascript implementation uses to encode the string into bytes. See https://stackoverflow.com/questions/30106476/using-javascripts-atob-to-decode-base64-doesnt-properly-decode-utf-8-strings

The first problem we want to solve is to first encode a javascript string into utf-8 byte sequence and then use base64 to encode the byte sequence into ascii string in browser environment.

The second problem is to inject such routine so that core is still no dependency.

louischan-oursky:
@kiootic Can you try to release a new alpha version? I would like to see if the instructions in README is clear enough for team members to do the release.

Issue #522, opened 2019-09-18
MFA


louischan-oursky:
- [x] Change the condition on when to refresh the access token
- [x] Ensure refresh can be retried
- [x] Implement MFA container
- [x] Handle bearer token transparently
- [x] Persist authn session token
- [x] Ensure mutually exclusiveness between (access token and refresh token) and authn session token
- [x] Expose authentication session in a beginner friendly way

PR #523, opened 2019-09-20
MFA

louischan-oursky:
connect #522

### DIFF
--- a/packages/skygear-core/src/container.ts
+++ b/packages/skygear-core/src/container.ts
@@ -14,6 +14,8 @@ import {
   CreateNewTOTPResult,
   ActivateTOTPResult,
   AuthenticateWithTOTPOptions,
+  CreateNewOOBOptions,
+  CreateNewOOBResult,
 } from "./types";
 import { BaseAPIClient, _removeTrailingSlash, encodeQuery } from "./client";
 import { SkygearError } from "./error";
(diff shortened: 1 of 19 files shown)


louischan-oursky:
@kiootic In case you forgot this PR :(

packages/skygear-core/src/error.ts
```diff
@@ -1,26 +1,95 @@
 import { JSONObject } from "./types";
 
+/**
+ * @public
+ */
+export const SkygearErrorNameUnexpectedError = "UnexpectedError" as const;
```
kiootic: Can we consolidate the error names in an object? e.g.
```typescript
export const SkygearErrorNames = {
    UnexpectedError: 'UnexpectedError',
    // ...
};
```

packages/skygear-core/src/container.ts
```diff
@@ -187,8 +187,18 @@ export class AuthContainer<T extends BaseAPIClient> {
...
+    // the access token in the request is invalid.
+    //
+    // If the request does not have access token at all,
+    // the server simply returns NotAuthenticated error without the header.
+    //
+    // Therefore, we have to keep the invalid token so that
+    // if refresh fails due to other reasons, the whole process
+    // can be retried.
+    //
+    // await this.parent.storage.delAccessToken(this.parent.name);
```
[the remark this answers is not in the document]
louischan-oursky: I kept it deliberately to warn us that they should not be there.

packages/skygear-core/src/error.ts
louischan-oursky: OK.

louischan-oursky:
@kiootic Updated.

Issue #519, opened 2019-09-16
defaultContainer and ES6 default export


louischan-oursky:
- [x] Use default export in commonjs and ES6
- [x] For script tag, merge the container and the namespace.

PR #524, opened 2019-09-23
defaultContainer and ES6 default export

louischan-oursky:
connect #519 

Since https://github.com/SkygearIO/skygear-SDK-JS/pull/523 added many examples, this PR need to rebase after it is merged.

### DIFF
--- a/packages/skygear-web/src/index.iife.ts
+++ b/packages/skygear-web/src/index.iife.ts
@@ -1,3 +1,14 @@
 import "core-js/features/promise";
 import "whatwg-fetch";
-export * from ".";
+
+import * as all from ".";
+// eslint-disable-next-line no-duplicate-imports
+import defaultContainer from ".";
+
+const merged = defaultContainer;
+for (const key of Object.keys(all)) {
+  // @ts-ignore
+  merged[key] = all[key];
+}
+
+export default merged;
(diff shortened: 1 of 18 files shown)


packages/skygear-web/src/index.iife.ts
```diff
@@ -1,3 +1,14 @@
...
-export * from ".";
+
+import * as all from ".";
+// eslint-disable-next-line no-duplicate-imports
+import defaultContainer from ".";
+
+const merged = defaultContainer;
+for (const key of Object.keys(all)) {
+  // @ts-ignore
+  merged[key] = all[key];
```
kiootic: Is it intentional that developer can use `skygear.default` to access the default container? i.e. `skygear === skygear.default.default.default.default`

louischan-oursky: It is a consequence of ES6 Module.

```typescript
export default a;
```

is equivalent to 

```typescript
export { a as default };
```

kiootic: I mean is this consequence intentional/desirable?

louischan-oursky: I would not say it is intentional because the intention here is to merge the container and the namespace so that using default container should appear more nature. I dont think it is a problem that they are circular.

PR #530, opened 2019-10-16
Allow ignoring error in logout

### LINKED ISSUE #529: Allow ignoring error in logout
ref SkygearIO/skygear-server#1063

kiootic:
fix #529

### DIFF
--- a/packages/skygear-core/src/container.ts
+++ b/packages/skygear-core/src/container.ts
@@ -306,8 +306,15 @@ export class AuthContainer<T extends BaseAPIClient> {
     );
   }
 
-  async logout(): Promise<void> {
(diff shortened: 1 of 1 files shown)


packages/skygear-core/src/container.ts
```diff
@@ -306,8 +306,14 @@ export class AuthContainer<T extends BaseAPIClient> {
     );
   }
 
-  async logout(): Promise<void> {
-    await this.parent.apiClient.logout();
+  async logout(force: boolean = false): Promise<void> {
```
louischan-oursky: Put `force` inside an object reads better.

```typescript
logout({ force: true});
// vs
logout(true);
```

PR #532, opened 2019-10-24
Asset Gear

louischan-oursky:


packages/skygear-core/src/imageprocessing.ts
```diff
@@ -0,0 +1,162 @@
...
+      ...options,
+    });
+    return this;
+  }
+
+  getName(): string {
+    return "pipeline";
+  }
+
+  getValue(): string {
```
kiootic: I think renaming `ImageProcessingPipelineBuilder` to `ImageProcessingPipeline` and `getValue` to `toString` would be better. This class is actually the pipeline itself, and the 'value' is the string representation of the pipeline.

packages/skygear-core/src/imageprocessing.ts
```diff
@@ -0,0 +1,162 @@
...
+          break;
+        default:
+          break;
+      }
+    }
+    return parts.join("/");
+  }
+
+  setToURLString(urlStr: string): string {
+    const fragmentIndex = urlStr.indexOf("#");
```
kiootic: Can we switch to using URL API? It seems it has more or less same level of support as fetch.

louischan-oursky: This is a trouble of supporting three platforms.

URL API is part of the Web API. Node.js has it but we have to import it from the `url` module. React Native does not have it at all.

I tried to find good URL and URLSearchParams polyfill but had no luck. Some of them only polyfill URL or URLSearchParams. If both are polyfilled, they polyfill the global which cannot be used by us because we are writing a library.

If we require the user to provide such polyfill in React Native, it is not a good user experience.

So I finally decided to implement `setToURLString` without using any library or API.

kiootic: Does [`whatwg-url`](https://github.com/jsdom/whatwg-url) fits the requirement?

louischan-oursky: It has many dependencies

kiootic: You're right. Let's keep it as is for now. For the record, I just tested in NodeJS and it does have a global `URL` class in version 10, so the problem would be on React Native.

packages/skygear-core/src/imageprocessing.ts
louischan-oursky: I think the Builder suffix better communicate the purpose.

For `toString`, I have thought about it when I wrote the spec. If it is named `toString`, should the return value be `pipeline=image/...` or `image/...`? Naming it toString would bring such ambiguity so I finally settled with calling it `getValue`

louischan-oursky:
ping @kiootic

PR #540, opened 2019-11-07
Fix not exporting validation typing & functions

kiootic:


packages/skygear-core/src/index.ts
```diff
@@ -6,6 +6,7 @@ export * from "./encoding";
 export * from "./storage";
 export * from "./url";
 export * from "./imageprocessing";
+export * from "./validation";
```
louischan-oursky: I always forgot to add new module too...

PR #544, opened 2019-11-27
Universal oauth flow

louischan-oursky:
connect https://github.com/SkygearIO/skygear-server/issues/1130


packages/skygear-web/src/container.ts
```diff
@@ -207,60 +243,49 @@ export class WebAuthContainer<T extends WebAPIClient> extends AuthContainer<T> {
...
-
-      const message = await messagePromise;
-      return decodeMessage(message);
-    } finally {
-      if (iframe) {
-        this.oauthResultObserver.unsubscribe();
-        document.body.removeChild(iframe);
-      }
+    const searchParams = new URLSearchParams(window.location.search.slice(1));
+    const result = searchParams.get("x-skygear-result");
```
kiootic: What do you think about erasing the parameter after reading it using `replaceState`? So that it would not leave potentially sensitive information in browser history.

packages/skygear-web/src/pkce.ts
```diff
@@ -0,0 +1,51 @@
+import { _encodeUTF8, _encodeBase64URLFromUint8Array } from "@skygear/core";
+
+function byteToHex(byte: number): string {
+  return ("0" + byte.toString(16)).substr(-2);
+}
+
+function numberArrayToUint8Array(arr: number[]): Uint8Array {
```
kiootic: We can use `Uint8Array.from(arr)` instead.

packages/skygear-web/src/pkce.ts
```diff
@@ -0,0 +1,51 @@
...
+  const promiseOrEvent = window.crypto.subtle.digest("SHA-256", bytes.buffer);
+  // eslint-disable-next-line
+  if (promiseOrEvent.then) {
+    // @ts-ignore
+    return promiseOrEvent.then((output: ArrayBuffer) => {
+      return new Uint8Array(output);
+    });
+  }
+  return new Promise((resolve, reject) => {
+    (promiseOrEvent as any).oncomplete = function(output: ArrayBuffer) {
```
kiootic: Let's add a comment here to indicate it's for support of IE11. (seems we haven't determine the supported browsers?)

packages/skygear-web/src/container.ts
louischan-oursky: I am afraid that it may break the application. The application may be using a library to manipulate history and does not expect the actual history to be modified directly.

packages/skygear-web/src/pkce.ts
louischan-oursky: IE does not have that static function

packages/skygear-web/src/pkce.ts
louischan-oursky: Sure. I will add it. I guess we should support IE 11. And by supporting IE 11, we should have supported many other browsers.

packages/skygear-web/src/container.ts
kiootic: I agree with your concern. I suppose we'll mention in the documentation that, the developer should erase the query parameters for better security.

PR #548, opened 2019-12-20
React Native oauth and Sign in with Apple

### LINKED ISSUE #547: Support OAuth and Sign in with Apple in React Native
- [x] Implement SecureRandomBytes and SHA256 on Android
- [x] Implement SecureRandomBytes and SHA256 on iOS
- [x] Support OAuth on iOS via [Linking.openURL](https://facebook.github.io/react-native/docs/linking#openurl)
- [x] Support OAuth on iOS >= 11 via [SFAuthenticationSession](https://developer.apple.com/documentation/safariservices/sfauthenticationsession?language=objc)
- [x] Support OAuth on iOS >= 12 via [ASWebAuthenticationSession](https://developer.apple.com/documentation/authenticationservices/aswebauthenticationsession)
- [x] Support OAuth on Android via [Linking.openURL](https://facebook.github.io/react-native/docs/linking#openurl)
- [x] Support Sign in with Apple on iOS by using UXModeManual

louischan-oursky:
fix #547

### DIFF
--- a/packages/skygear-core/src/base64.ts
+++ b/packages/skygear-core/src/base64.ts
@@ -1,9 +1,5 @@
 import { _encodeUTF8 } from "./utf8";
-
-interface ByteArray {
-  [index: number]: number;
-  length: number;
-}
+import { ByteArray } from "./types";
 
 // https://developer.mozilla.org/en-US/docs/Web/API/WindowBase64/Base64_encoding_and_decoding
 function uint6ToB64(nUint6: number) {
@@ -89,6 +85,6 @@ export function _encodeBase64FromString(input: string): string {
 /**
  * @internal
  */
-export function _encodeBase64URLFromUint8Array(bytes: Uint8Array): string {
+export function _encodeBase64URLFromByteArray(bytes: ByteArray): string {
   return base64URLEncArr(bytes);
 }
--- a/packages/skygear-core/src/base64.ts
+++ b/packages/skygear-core/src/base64.ts
@@ -88,3 +88,50 @@ export function _encodeBase64FromString(input: string): string {
 export function _encodeBase64URLFromByteArray(bytes: ByteArray): string {
   return base64URLEncArr(bytes);
 }
+
+/**
+ * @internal
+ */
+export function _windowAtob(string: string): string {
+  // This is borrowed from
+  // https://github.com/MaxArt2501/base64-js/blob/master/base64.js
+  const b64 =
+    "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/=";
+  // Regular expression to check formal correctness of base64 encoded strings
+  // eslint-disable-next-line no-useless-escape
+  const b64re = /^(?:[A-Za-z\d+\/]{4})*?(?:[A-Za-z\d+\/]{2}(?:==)?|[A-Za-z\d+\/]{3}=?)?$/;
+  // atob can work with strings with whitespaces, even inside the encoded part,
+  // but only \t, \n, \f, \r and ' ', which can be stripped.
+  string = String(string).replace(/[\t\n\f\r ]+/g, "");
+  if (!b64re.test(string))
+    throw new TypeError(
+      "Failed to execute 'atob' on 'Window': The string to be decoded is not correctly encoded."
+    );
+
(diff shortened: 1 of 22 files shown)


packages/skygear-react-native/android/src/main/java/io/skygear/reactnative/SGSkygearReactNativeModule.java
```diff
@@ -0,0 +1,134 @@
...
+            if (this.openURLPromise != null) {
+                this.openURLPromise.reject(e);
+                this.openURLPromise = null;
+            }
+        }
+    }
+
+    @ReactMethod
+    public void signInWithApple(String urlString, Promise promise) {
+        promise.reject("EUNSPECIFIED", "Android does not support Sign in with Apple");
```
kiootic: I think for RN Android, we'll open chrome and use default OIDC flow?

packages/skygear-react-native/ios/SGSkygearReactNative.m
```diff
@@ -0,0 +1,353 @@
...
+  }
+  return YES;
+}
+
+RCT_EXPORT_METHOD(openURL:(NSURL *)url
+                   scheme:(NSString *)scheme
+                  resolve:(RCTPromiseResolveBlock)resolve
+                   reject:(RCTPromiseRejectBlock)reject)
+{
+    self.openURLResolve = resolve;
```
kiootic: I suppose this function does not support reentrancy on iOS? We should reject the previous promise with some error, rather than leaving it hanging.

packages/skygear-react-native/ios/SGSkygearReactNative.m
```diff
@@ -0,0 +1,353 @@
...
+    self.provider = provider;
+    self.controller = controller;
+    self.state = state;
+    self.nonce = nonce;
+    self.signInWithAppleResolve = resolve;
+    self.signInWithAppleReject = reject;
+
+    [controller performRequests];
+  } else {
+    reject(RCTErrorUnspecified, @"Sign in with Apple requires iOS 13", nil);
```
kiootic: Do we want to support older version through standard OIDC flow/JS SDK?

packages/skygear-react-native/ios/SGSkygearReactNative.m
louischan-oursky: Yes, you're right. It does not support reentrancy. I do not think there is a valid use case to login concurrently.

packages/skygear-react-native/android/src/main/java/io/skygear/reactnative/SGSkygearReactNativeModule.java
louischan-oursky: I wanted to make the API simpler. `loginApple` always means iOS AuthenticationServices  flow while `loginOAuthProvider` always means OIDC flow.

packages/skygear-react-native/ios/SGSkygearReactNative.m
louischan-oursky: explained in a previous comment.

packages/skygear-core/src/base64.ts
```diff
@@ -89,6 +85,53 @@ export function _encodeBase64FromString(input: string): string {
...
+export function _encodeBase64URLFromByteArray(bytes: ByteArray): string {
   return base64URLEncArr(bytes);
 }
+
+/**
+ * @internal
+ */
+export function _windowAtob(string: string): string {
+  // This is borrowed from
+  // https://github.com/MaxArt2501/base64-js/blob/master/base64.js
```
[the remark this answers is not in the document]
louischan-oursky: I prefer we include the license of the borrowed code. Where should we put it?

packages/skygear-react-native/ios/SGSkygearReactNative.m
kiootic: If so, we may want to have a more descriptive error message, e.g. "use standard OIDC flow instead of this function".

louischan-oursky:
Ready for review.

louischan-oursky:
@kiootic Any further concerns?

kiootic:
No more concerns from me, just waiting for the spec to be merged first. I think we can merge this first?

PR #553, opened 2020-01-08
Add isManuallyVerified flag

kiootic:


packages/skygear-core/src/storage.test.ts
```diff
@@ -36,6 +36,7 @@ describe("ContainerStorage", () => {
       id: "userid",
       createdAt: new Date(0),
       lastLoginAt: new Date(0),
+      isManuallyVerified: true,
```
louischan-oursky: I know the value does not really matter. But why not setting it to false?

kiootic: No particular reason, just want to use some more varieties of value for testing.

Issue #554, opened 2020-03-16
Support using auth ui to login with OIDC flow


carmenlau:
- [x] web
- [x] react-native

PR #556, opened 2020-03-19
Web authorize with oidc flow

carmenlau:
require #557 
connect #554 

To try the example with other provider like `keycloak`, we can configure to provide keycloak endpoint in auth endpoint, e.g.:
```
SKYGEAR_APP_ENDPOINT=ANY_VALID_ENDPOINT \
SKYGEAR_AUTH_ENDPOINT=http://localhost:8080/auth/realms/master \
SKYGEAR_CLIENT_ID=CLIENT_ID \
# SKYGEAR_IS_THIRD_PARTY_APP=true \
npm run example
```

TODO:
- Update error handling, should have OAuthError should be an error object instead interface

TBD:
- ~~The current flow use response of userinfo endpoint to construct the user, we should decide what attributes will be return from the userinfo endpoint, or we should change some of the user object attributes optional~~
- ~~The new SDK use oidc token endpoint to renew token, we assume that user is using the new server with new sdk.~~
- But I am not sure how old sdk works with new server, if the new gateway behavior is always proxy the request and no longer have header `x-skygear-try-refresh-token`? Or we don't need to take care this? As the old app that we need to migrate is using cookie transport?

Remarks:
- old api auth response should update to return expires_in, otherwise sdk will not call refresh token

### DIFF
--- a/packages/skygear-core/src/client.ts
+++ b/packages/skygear-core/src/client.ts
@@ -100,6 +100,8 @@ export abstract class BaseAPIClient {
   userAgent?: string;
   getExtraSessionInfo?: () => Promise<JSONObject | null>;
 
+  private config?: _OIDCConfiguration;
+
   constructor() {
     this.apiKey = "";
     this.appEndpoint = "";
@@ -911,18 +913,19 @@ export abstract class BaseAPIClient {
    * @internal
    */
   async _fetchOIDCConfiguration(): Promise<_OIDCConfiguration> {
-    return this._fetchOIDCRequest(
-      `${this.authEndpoint}/.well-known/openid-configuration`
-    );
+    if (!this.config) {
+      this.config = (await this._fetchOIDCRequest(
+        `${this.authEndpoint}/.well-known/openid-configuration`
+      )) as _OIDCConfiguration;
+    }
+    return this.config;
   }
 
   /**
    * @internal
    */
-  async _oidcTokenRequest(
-    tokenEndpoint: string,
-    req: _OIDCTokenRequest
-  ): Promise<_OIDCTokenResponse> {
+  async _oidcTokenRequest(req: _OIDCTokenRequest): Promise<_OIDCTokenResponse> {
+    const config = await this._fetchOIDCConfiguration();
     const query = encodeQuery([
       ["grant_type", req.grant_type],
       ["code", req.code],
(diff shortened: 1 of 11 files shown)


packages/skygear-core/src/container.ts
```diff
@@ -822,6 +823,86 @@ export class MFAContainer<T extends BaseAPIClient> {
...
+    const stateQuery = options.state ? `state=${options.state}` : "";
+    if (this.isThirdParty) {
+      const codeVerifier = await this._setupCodeVerifier();
+      await this.parent.parent.storage.setOIDCCodeVerifier(
+        this.parent.parent.name,
+        codeVerifier.verifier
+      );
+
+      return (
+        `${config.authorization_endpoint}?` +
```
louischan-oursky: Use `encodeQuery` to encode query properly.

kiootic:
> The UserInfo endpoint minimally should return the decoded ID token. If we want to return User, we need to introduce our own claims in the ID token. 

How about using `/me` API as well? I remembered we discussed on scopes and claims, and we don't want to introduce our own scopes at this stage.

> I do not remember if c<redacted> is using cookie or header.

I remembered last time we checked with developer to confirm that they're using tokens (i.e. header).

louischan-oursky:
Using the `/me` means Auth API must be enabled for SDK to work properly.

carmenlau:
Update the sdk based on yesterday discussion, now I only assume the userinfo endpoint is returning the skygear's claims. Since keycloak doesn't have those claims, will test the sdk again when the userinfo endpoint is ready. 🙇🏻‍♀️

The updated spec for your reference: 
https://github.com/SkygearIO/features/pull/425/commits/56576c3defb96cf31770486f25e8d14018aa9674
https://github.com/SkygearIO/features/pull/425/commits/d81a17b455c8415dccf1d9cead8f3e96ee54c7e8

packages/skygear-core/src/client.ts
```diff
@@ -857,4 +884,90 @@ export abstract class BaseAPIClient {
...
+   */
+  async _oidcUserInfoRequest(accessToken?: string): Promise<AuthResponse> {
+    const headers: { [name: string]: string } = {};
+    if (accessToken) {
+      headers["authorization"] = `bearer ${accessToken}`;
+    }
+    const config = await this._fetchOIDCConfiguration();
+    const userinfo = await this._fetchOIDCRequest(config.userinfo_endpoint, {
+      method: "GET",
+      headers: headers,
```
louischan-oursky: Add 

```typescript
mode: "cors",
credentials: "include"
```

because we sometimes expect cookie to be included for the userinfo endpoint

carmenlau:
Added one more commit for fixing first party app current user not persist problem 2cc83ae

PR #558, opened 2020-03-22
React native authorize with oidc flow

carmenlau:
require #556 
connect #554

### DIFF
--- a/packages/skygear-react-native/android/src/main/java/io/skygear/reactnative/SGSkygearReactNativeModule.java
+++ b/packages/skygear-react-native/android/src/main/java/io/skygear/reactnative/SGSkygearReactNativeModule.java
@@ -59,6 +59,21 @@ public class SGSkygearReactNativeModule extends ReactContextBaseJavaModule {
     }
 
     @ReactMethod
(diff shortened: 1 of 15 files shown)


packages/skygear-react-native/android/src/main/java/io/skygear/reactnative/SGSkygearReactNativeModule.java
```diff
@@ -58,6 +58,25 @@ public void sha256String(String input, Promise promise) {
         }
     }
 
+    @ReactMethod
+    public void openURL(String urlString, Promise promise) {
```
louischan-oursky: The current implementation uses normal webview to open authorization URL and normal URL. However, we should be using [Chrome CustomTabs](https://github.com/SkygearIO/features/pull/425/files#diff-9888891478832f53b6dfd9966dabd74fR143). This is also [the best practice](https://tools.ietf.org/html/rfc8252#appendix-B.2). Let's open an issue to address this.

carmenlau: Sure, opened issue #559

PR #564, opened 2020-03-26
Authui logout

carmenlau:
refs #554 

Discussion:
For first party app, logout function will bring user to end_session_endpoint and redirect back to the app after logout. At this moment, the current user state of SDK is not cleared yet.  Do we need something like `finishLogout`? `end_session_endpoint` will only redirect back to the app only when logout success, so the only `finishLogout` function will do is clear session, and it will be used by first party app only.  Or maybe we should only expose the `clearSession` function for the user, so user should call `clearSession` in their after logout screen? thought?

### DIFF
--- a/packages/skygear-core/src/types.ts
+++ b/packages/skygear-core/src/types.ts
@@ -702,6 +702,7 @@ export interface _OIDCConfiguration {
   claims_supported: string;
   code_challenge_methods_supported: string;
   revocation_endpoint: string;
+  end_session_endpoint: string;
 }
 
 /**
--- a/packages/example/src/authui.html
+++ b/packages/example/src/authui.html
@@ -46,6 +46,22 @@
       <textarea id="finishAuthorizationResult" class="form-control" readonly=""></textarea>
     </div>
   </form>
+  <form class="container" onsubmit="onSubmitLogout(event)">
+    <div class="form-group">
+      <h1>Logout</h1>
+    </div>
+    <div class="form-group">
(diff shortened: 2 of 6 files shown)


kiootic:
- We'll need a clear session function anyways, since when the app received a 401 from backend, it'll want to clear the session and try again (in case of revoked session).
- I'd expect the first thing to call after configuring the SDK would be `me()` (or equivalent), since the SDK cannot guarantee the user state on startup (maybe revoked, expired). Therefore there should be no need to call any 'after logout' function, since the user state should be refreshed with every page load.

louischan-oursky:
If it is first party app, the single source of truth should be cookie. We should not persist user and identity. After calling `.configure()`, the developer should call `/oauth2/userinfo` to get the correct state.

If it is third party app, the single source of truth should be the refresh token. We can persist user and identity.

packages/skygear-core/src/types.ts
```diff
@@ -702,6 +702,7 @@ export interface _OIDCConfiguration {
   claims_supported: string;
   code_challenge_methods_supported: string;
   revocation_endpoint: string;
+  end_session_endpoint: string;
```
louischan-oursky: Some fields here have wrong types. `scopes_supported`, `response_types_supported`, `grant_types_supported`, `subject_types_supported`, `id_token_signing_alg_values_supported`, `claims_supported`, `code_challenge_methods_supported`. I think we can just remove the fields we do not use in the SDK. That being said, just keep `*_endpoint` fields.

PR #567, opened 2020-03-27
Update example instructions

### LINKED ISSUE #561: Update README to mention localhost DNS resolver is required
For the example to work

The example has to be at `http://app.localhost:9999`
The server has to be at `http://accounts.app.localhost:8081`

Otherwise the setup does not match the assumption of first party app.

louischan-oursky:
fix #561


louischan-oursky:
ping

PR #571, opened 2020-04-15
Set response mode to query_html_redirect for ios version < 13.4

carmenlau:
To work around ios redirect bug when using auth ui to sign in with app in ios

refs skygeario/skygear-server#1334


carmenlau:
Decided to use js and meta to do the direct for all case, so don't need to specify response mode.

refs https://github.com/SkygearIO/skygear-server/pull/1335

PR #576, opened 2020-05-08
Add anonymous user APIs

kiootic:


packages/skygear-react-native/android/src/main/java/io/skygear/reactnative/SGSkygearReactNativeModule.java
```diff
@@ -117,6 +131,65 @@ public void getCredentialStateForUserID(String userID, Promise promise) {
...
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
+            promise.reject("EUNSPECIFIED", "Anonymous authentication is not supported on Android before version M.");
+            return;
+        }
+
+        try {
+            if (kid == null) {
+                kid = UUID.randomUUID().toString();
+            }
+            Map<String, Object> header = new HashMap<>();
```
louischan-oursky, line 145: On android, we need to use WritableMap and WritableArray to write JavaScript object and array respectively. Unlike iOS, native Java map and array types do not work :(

This page contains some examples https://reactnative.dev/docs/native-modules-android

louischan-oursky:
The ID token still have identity and the SDK cannot decode it. (unknown identity type). We are going to remove the identity from ID token, right?

kiootic:
> The ID token still have identity and the SDK cannot decode it. (unknown identity type). We are going to remove the identity from ID token, right?

I think so. We should update the existing issue (SkygearIO/skygear-server#1370) for planned breaking changes.

kiootic:
@louischan-oursky @carmenlau here is some sample code for testing RN SDK:
```ts
  const doAuth = useCallback(async () => {
    const {user} = await container.auth.authenticateAnonymously();
    console.log(user);
  }, []);

  const doPromote = useCallback(async () => {
    const result = await container.auth.promoteAnonymousUser({
      redirectURI: "rntest://callback"
    });
    console.log(result);
  }, []);
```

louischan-oursky:
It works fine on iOS simulator. I tested it with Android emulator Pixel 2 with API Level 29. It fails at a native method in SGSkygearReactNativeModule.java:187 with the root exception `android.security.KeyStoreException: Incompatible padding mode`. Any idea?

kiootic:
Strange, I remember fixing both this and the `WritableMap` issue before. Seems the changes somehow got lost. I'll fix it tomorrow then.

PR #580, opened 2020-05-15
Derive rollup external from package.json

### LINKED ISSUE #578: Fix react native bundle warning
Rollup complains about `react-native` being unresolved. We should mark all `peerDependencies` as externals.

louischan-oursky:
fix #578

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,18 +1,18 @@
-const promisify = require("util").promisify;
-const fs = require("fs");
-
-const ClosureCompiler = require("google-closure-compiler").compiler;
-const tmp = require("tmp");
-
+import { readFileSync, writeFile } from "fs";
+import { promisify } from "util";
 import babel from "rollup-plugin-babel";
 import commonjs from "rollup-plugin-commonjs";
 import json from "rollup-plugin-json";
 import resolve from "rollup-plugin-node-resolve";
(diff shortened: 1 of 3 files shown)


rollup.config.js
```diff
@@ -89,6 +89,53 @@ function makeBabelExternal(id) {
...
+
+  const peerDeps = Object.keys(nodePackageJSON["peerDependencies"] || {});
+  if (peerDeps.length > 0) {
+    throw new Error(
+      "@skygear/node-client should not have any peerDependencies"
+    );
+  }
+
+  const deps = Object.keys(nodePackageJSON["dependencies"] || {});
+  const builtins = getBuiltins();
```
kiootic, line 107: I suppose this returns a list of built-in modules of NodeJS. I don't think any of them would ever appears in dependencies on package.json, therefore the check should be redundant?

louischan-oursky: Yes, you are right. They wont appear in `dependencies` but it is perfectly fine to use built-in modules of Node in `@skygear/node-client`. The reason why we have `builtins` here is to write the `external` function. The `external` of node has two jobs

1. If the module (`id`) appear in `dependencies`, tell rollup that it is external.
2. Otherwise if the module (`id`) is a node builtin, tell rollup that is is external.

PR #581, opened 2020-05-18
Make content-type required in react native

### LINKED ISSUE #579: React native upload asset fails in Android
It fails in Android but Ok in iOS

- Android version: 10
- React native version: 0.62
- SDK version: 2.3.0
- file type: image

louischan-oursky:
fix #579

### DIFF
--- a/packages/skygear-react-native/src/index.ts
+++ b/packages/skygear-react-native/src/index.ts
@@ -72,8 +72,15 @@ async function uploadForm(
       form.append(name, value);
     }
   }
+
+  const contentType = (req.headers || {})["content-type"];
+  if (contentType == null) {
+    throw new Error("content-type is required to upload asset");
+  }
+
   form.append("file", {
     uri,
+    type: contentType,
     name: "filename",
   } as any);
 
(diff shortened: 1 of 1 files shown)


packages/skygear-react-native/src/index.ts
```diff
@@ -72,8 +72,15 @@ async function uploadForm(
       form.append(name, value);
     }
   }
+
+  const contentType = (req.headers || {})["content-type"];
```
kiootic, line 76: This is a breaking change on RN SDK API.

I see you mentioned 'On iOS, content type sometimes can be inferred automatically' in #579, do you know when would the inference fail?

louischan-oursky: My previous statement is wrong. On iOS, [content-type is not required](https://github.com/facebook/react-native/blob/master/Libraries/Network/RCTNetworking.mm#L114). So the inference is done by the server.

louischan-oursky: The expected usage is the developer provide the correct content-type so when the asset is accessed, the content-type is set correctly. The problem is that the android implementation requires content-type to be set at client side.

kiootic: Sorry I don't quite understand. Do you mean this?
- On iOS, `Content-Type` does not need to be provided at client side; it would be inferred at server-side.
- On Android, `Content-Type` is needed at client side; server side does not infer the content type.

louischan-oursky: Yes.

PR #582, opened 2020-05-18
Remove MFA APIs

kiootic:
ref SkygearIO/skygear-server#1426


kiootic:
Yes, since the identity models are not yet updated in SDK. This would be done in another PR after removing current identity from server.

For now you can test it by removing the `decodeIdentity` calls.

PR #583, opened 2020-05-19
Update identity handling

kiootic:
ref SkygearIO/skygear-server#1424

### DIFF
--- a/packages/skygear-core/src/types.ts
+++ b/packages/skygear-core/src/types.ts
@@ -53,7 +53,18 @@ export interface User {
 /**
  * @public
  */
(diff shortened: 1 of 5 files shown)


packages/skygear-core/src/types.ts
```diff
@@ -53,7 +53,20 @@ export interface User {
...
+}
+
+/**
+ * @public
+ */
+export interface IdentityClaims {
+  [claim: string]: unknown;
+  email?: string;
+  phone?: string;
+  username?: string;
```
louischan-oursky, line 68: Just keep `email` first? Btw `phone` in OIDC standard claims is called `phone_number` actually.

PR #585, opened 2020-05-25
Fix decode query should ignore fragment

carmenlau:
fix SkygearIO/skygear-server#1448


cychiuae:
:adore2x:

packages/skygear-core/src/url.ts
```diff
@@ -43,11 +43,13 @@ export function decodeQueryComponent(s: string): string {
 /**
  * @public
  */
-export function decodeQuery(query?: string): [string, string][] {
-  if (!query) {
+export function decodeQuery(path?: string): [string, string][] {
```
louischan-oursky, line 46: Making `decodeQuery` to support fragment would complicate the function. I suggest we ensure the caller only pass valid query to `decodeQuery`. Currently the only caller just split the URL by `?` so we have this bug. Maybe we can introduce a new function to split URL into 3 parts

```typescript
// splitQueryFragment splits url into 3 parts, returned as a 3-tuple.
// If query or fragment is absent, the corresponding part is an empty string.
function splitQueryFragment(url: string): [string, string, string] {
  // ...
}
```

kiootic: I remembered we discussed on polyfill for `URL` last time: https://github.com/SkygearIO/skygear-SDK-JS/pull/532#discussion_r339457614

I just searched a bit more and seems core-js has a pure version of URL polyfill: `core-js-pure/features/url`. It doesn't pollute global namespace, so we may look into using it.

carmenlau: So we may add `core-js-pure` as `skygear-core` dependency, looks good to me

louischan-oursky: Note that we want to bundle the polyfill so it should be a dev dependency and the consumer of the SDK does not need to install it. We also need to verify we only bundle the URL polyfill and nothing else.

carmenlau:
Found that the updated version with core-js doesn't work in react-native, the pr is not ready yet, investigating.

carmenlau:
> Found that the updated version with core-js doesn't work in react-native, the pr is not ready yet, investigating.

The mentioned problem is addressed in 390a3e9, and the pr is ready for review.

carmenlau:
Updated!

packages/skygear-core/src/url.test.ts
```diff
@@ -1,9 +1,4 @@
-import {
-  encodeQueryComponent,
-  encodeQuery,
-  decodeQueryComponent,
-  decodeQuery,
-} from "./url";
+import { encodeQueryComponent } from "./url";
```
kiootic, line 1: I suppose we can replace the logic in image processing URL builder, so this function can be deleted as well.

packages/skygear-core/src/core-js.ts
```diff
@@ -1,3 +1,7 @@
 declare module "core-js-pure/features/url" {
   export default URL;
 }
+
+declare module "core-js-pure/features/url-search-params" {
+  export default URLSearchParams;
+}
```
louischan-oursky, line 7: Sorry, this file should be called `core-js.d.ts` to signify it is a declaration file. I should have told you this yesterday. My bad.

packages/skygear-core/src/url.test.ts
carmenlau: Updated! So great that we have test in image processing!XD @louischan-oursky

packages/skygear-core/src/core-js.ts
carmenlau: Updated in 91f4bb6. I updated the `clean-decl` command too, see if it looks ok?

PR #586, opened 2020-05-26
Cancel

### LINKED ISSUE #575: How to detect cancel in authorization flow
With https://github.com/SkygearIO/skygear-SDK-JS/pull/574 if the user cancels the promise never settle. In some cases the developer may want to detect the cancel case.

Here is an idea to detect the cancel case. Pseudo code only!

```typescript
import { AppState } from "react-native";
import EventEmitter from "react-native/Libraries/vendor/emitter/EventEmitter";

class CancelWatcher extends EventEmitter {
  constructor(opts) {
    super();
    this.opts = opts;
  }
  watch() {
    if (this.watching) { return; }
    AppState.addEventListener("change", this.onAppStateChange);
  }
  onAppStateChange = (nextAppState) => {
    if (nextAppState === "active") {
      this.timeoutToken = setTimeout(this.onTimeout, this.opts.delayMillis);
    }
  };
  dispose() {
    if (this.timeoutToken) {
      clearTimeout(this.timeoutToken);
      this.timeoutToken = null;
    }
    if (this.watching) {
      AppState.removeEventListener("change", this.onAppStateChange);
      this.watching = false;
    }
  }
  onTimeout = () => {
    this.dispose();
    this.emit("cancel");
  };
}

async function foobar() {
  const cancelWatcher = new CancelWatcher({ delayMillis: 1000});
  cancelWatcher.addEventListener("cancel", function() {
    // Do something on cancel...
  });
  try {
    cancelWatcher.watch();
    await skygear.auth.authorize();
  } finally {
    cancelWatcher.dispose();
  }
}
```

louischan-oursky:
fix #575


packages/skygear-react-native/android/src/main/java/io/skygear/reactnative/SGSkygearReactNativeModule.java
```diff
@@ -188,6 +189,32 @@ public void signAnonymousToken(String kid, String data, Promise promise) {
...
 
+    @Override
+    public void onActivityResult(Activity activity, int requestCode, int resultCode, Intent data) {
+        if (this.openURLPromise == null) {
+            return;
+        }
+
+        if (requestCode == REQUEST_CODE_AUTHORIZATION) {
+            if (resultCode == Activity.RESULT_CANCELED) {
+                this.openURLPromise.reject("CANCEL", "CANCEL");
```
carmenlau, line 200: The detection looks good. So now for developer who want to detect cancel they will need to check `err.message === "CANCEL"`? Will we consider having SDKError class per our discussion?
Btw one minor different between android and ios, the err.code in android is `CANCAL` and ios is `EUNSPECIFIED`. We may want it to be consistent

PR #591, opened 2020-06-04
Authgear

louischan-oursky:


kiootic (commented):
I remember from the previous discussion that, we can just start a new repo and rewrite the SDK, since so much would need to be changed?

louischan-oursky:
I want to keep the history. Starting from scratch has some overhead.

kiootic (commented):
We may as well delete `.travis.yaml` and `CHANGELOG`?

louischan-oursky:
updated.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,32 @@
...
+    runs-on: ubuntu-18.04
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-node@v1
+      with:
+        node-version: "12"
+    - run: npm ci
+    - run: npm run lerna bootstrap
+    - run: npm run format
+      if: ${{ always() }}
```
kiootic: I suppose `!cancelled()` would be better.

louischan-oursky: updated.

kiootic (approved):

